Issue #116, opened 2013-08-23
unicode errors when captured output includes non-ascii characters


jpellerin:
Some bad unicode hygiene in the output plugin, most likely.

PR #150, opened 2014-02-12
Encode according to stream encoding

jpellerin:
Here's another stab at #116/#130 -- still missing a test that shows the failure case, but this is working for me in practice, in combination with the other unicode/ascii fix I just merged.


katrinabrock:
#346 merged these changes

PR #232, opened 2015-01-04
Project directory inside package (fixes #161)

### LINKED ISSUE #161: nose2 cant import tests from package
I have set up a virtual env and have a test structure that looks something like this:
<path>/test_package/__init__.py
                                 test_foo.py:test_bar

test_bar is a really simple test, it just passes.
When I stand in "<path>/test_package" and run "nose2" I get an error:
"ImportError: No module named test_package.test_foo"
But it works with nosetests.
And if I stand in "<path>" and run "nose2 test_package" it works, and it also works if I remove the __init__.py

I know that the collection of tests work a little different than in nosetests, but I think that it would be good to make this work since it would make it easier to create a test package and run it.

kurniliya:
Fixes #161. The fix I prepared is to ensure that necessary path ('/' in the reported case) is in `sys.path` before attempting to import a test module '/test_package/test_foo.py' as `test_package.test_foo` during discovery.
I shall be happy to work more on the patch shall there be comments/suggestions/etc.


thedrow:
This looks great.
Can you please add a unit test to ensure_importable so we'll make sure that this issue won't regress?

thedrow:
Thanks. It will be available on the next release.

PR #236, opened 2015-02-22
add missing Makefile for docs in sdist

alunduil:
When packaging docs as part of the Gentoo ebuild I found the Makefile
for documentation wasn't included in the sdist.  This simple patch
ensures that it's included.


thedrow:
Sorry about that :)

PR #238, opened 2015-03-04
defensive: dont assume a traceback object, fixes #237

### LINKED ISSUE #237: TypeError when using ext. log handling
My tests handle exception logging elsewhere, and nose2 now fails in util.format_traceback() since sys.exc_info() now has None as traceback member (last)

sabaini:


sabaini:
Actually this only half a fix, let me come up with something better ...

PR #245, opened 2015-05-27
Restore exception context when passed in loader's _makeFailedTest method

### LINKED ISSUE #48: Loader errors throw away tracebacks and exception detail
For instance if a generator test throws an AttributeError, the details of the attribute error are lost and the user only sees "AttributeError" -- not very helpful.

dlax:
Closes #48.

### DIFF
--- a/nose2/loader.py
+++ b/nose2/loader.py
@@ -7,6 +7,8 @@
 import logging
 import traceback
 
+import six
+
 from nose2 import events
 from nose2.compat import unittest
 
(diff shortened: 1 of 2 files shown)


dlax:
Hmm, this does not actually fixes #48 completly since the exception context is not forwarded down to the test failure. Will work further on this...

dlax:
Pushed a new version.

dlax:
Another version hopefully suitable for Python 3...

thedrow:
Can you ensure all branches are covered by tests?

Issue #235, opened 2015-02-22
failing test (nose2.tests.functional.test_coverage.TestCoverage) in 0.5.0


alunduil:
From what I can tell the output expected for coverage in the following test no longer matches expectations (I'm probably just not fully aware of what's happening).  I'm working on packaging this for Gentoo and could be missing a dependency as well.  Any guidance would be appreciated.

Failing test:

```
======================================================================
FAIL: test_run (nose2.tests.functional.test_coverage.TestCoverage)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./nose2/tests/functional/test_coverage.py", line 19, in test_run
    stderr=os.path.join('lib', 'mod1').replace('\\', r'\\') + STATS)
  File "./nose2/tests/_common.py", line 64, in assertTestRunOutputMatches
    testf(util.safe_decode(cmd_stderr), stderr)
AssertionError: Regex didn't match: 'lib/mod1           8      5    38%' not found in 'test1 (test_coverage.TestLib) ... ok\n\n----------------------------------------------------------------------\nRan 1 test in 0.001s\n\nOK\n'

-------------------- >> begin captured logging << --------------------
nose2.main: DEBUG: logging initialized 30
nose2.session: DEBUG: Loading plugin modules: {'nose2.plugins.logcapture', 'nose2.plugins.result', 'nose2.plugins.loader.generators', 'nose2.plugins.loader.testclasses', 'nose2.plugins.buffer', 'nose2.plugins.loader.discovery', 'nose2.plugins.loader.testcases', 'nose2.plugins.coverage', 'nose2.plugins.loader.functions', 'nose2.plugins.failfast', 'nose2.plugins.loader.loadtests', 'nose2.plugins.loader.parameters', 'nose2.plugins.debugger', 'nose2.plugins.dundertest'}
nose2.session: DEBUG: Plugin is available: <class 'nose2.plugins.logcapture.LogCapture'>
nose2.session: DEBUG: Plugin is available: <class 'nose2.events.Plugin'>
nose2.session: DEBUG: Plugin is available: <class 'nose2.plugins.result.ResultReporter'>
nose2.session: DEBUG: Register active plugin <nose2.plugins.result.ResultReporter object at 0x7fb780bd11d0>
nose2.session: DEBUG: Register method startTest for plugin <nose2.plugins.result.ResultReporter object at 0x7fb780bd11d0>
[...]

PR #255, opened 2015-11-10
Fix "TypeError: exceptions must derive from BaseException" in loader error handling.

leth:
Passing on `sys.exc_info()` results in an error in `PluggableTestLoader._makeFailedTest` because tuple is not a subclass of `BaseException`.

### DIFF
--- a/nose2/plugins/loader/generators.py
+++ b/nose2/plugins/loader/generators.py
@@ -182,12 +182,11 @@ class Generators(Plugin):
                 method = functools.update_wrapper(method, func)
                 setattr(instance, method_name, method)
                 yield instance
-        except:
-            exc_info = sys.exc_info()
+        except Exception as e:
             test_name = '%s.%s.%s' % (testCaseClass.__module__,
                                       testCaseClass.__name__,
                                       name)
-            yield event.loader.failedLoadTests(test_name, exc_info)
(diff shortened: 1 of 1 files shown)


thedrow:
can you fix the failing test? the master's build is green so that's caused by your change.

leth:
I don't think it's my change. Master was last built 4 months ago. There's been a major release of the coverage package since then.

Digging into the failure, the regex `lib/mod1           8      5    38%` fails to match the output below:

```
'test1 (test_coverage.TestLib) ... ok

Ran 1 test in 0.002s

OK
----------- coverage: platform linux, python 3.4.2-final-0 -----------
Name              Stmts   Miss  Cover
-------------------------------------
lib/__init__.py       0      0   100%
lib/mod1.py           8      5    38%
-------------------------------------
TOTAL                 8      5    38%
```

leth:
I believe this is Issue #235

leth:
Thanks! I raised another PR which should fix the build.

On 10 November 2015 18:39:06 GMT+00:00, Omer Katz <notifications@github.com> wrote:
>Merged #255.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/nose-devs/nose2/pull/255#event-460441039

-- 
Sent from my phone. Please excuse my brevity.

Issue #215, opened 2014-07-24
Such DSL: setup runs out of order when using @it.uses


lukeorland:
I have an example here ([gist]) of an internal layer's tests being run before its setup when the setup is defined in the `Layer` class and then accessed via `@it.uses(Layer)`.

[gist]: https://gist.github.com/lukeorland/d4534a20241d7b5280b0

little-dude:
Fixed. Will be in next release.

PR #262, opened 2016-02-13
add test for out of order setups with such

little-dude:
trying to reproduce https://github.com/nose-devs/nose2/issues/215 with CI, since I couldn't reproduce locally with python 2.7 and python 3.5.

@lukeorland : I shamelessly copy-pasted your gist, so please let me know if you want to be the author of the commit or if you want to do the PR yourself.


lukeorland:
Nice PR--I don't need to be owner. Thanks for asking.

little-dude:
Thanks :)

Seems that travis doesn't run these tests...

PR #267, opened 2016-02-16
refactor layers

little-dude:
This is an attempt to fix https://github.com/nose-devs/nose2/issues/215

The problem is that when sorting the layers, we treat layer mixins and layer parents the same way. Mixins cannot be treated like regular layers, because they don't always have parents. Instead, mixins should be inserted between the layer the parent layer and the layers they belong to, whatever their parents are.

To take the example given in the issue, here is how layers are currently resolved:
```
no layer ->
  <class 'test_such_timing.Layer1'> ->
    <class 'test_such_timing.A system with a fast initial setup layer:layer'> ->
     test 0000: should not have obj initialized (test_such_timing.A system with a fast initial setup layer)
      <class 'test_such_timing.having a second slow setup layer:layer'> ->
       test 0000: should have obj initialized (test_such_timing.having a second slow setup layer)
      <- <class 'test_such_timing.having a second slow setup layer:layer'>
    <- <class 'test_such_timing.A system with a fast initial setup layer:layer'>
  <- <class 'test_such_timing.Layer1'>
  <class 'test_such_timing.Layer2'> ->
  <- <class 'test_such_timing.Layer2'>
<- no layer
```

`Layer2` is a mixin of the `having a second slow setup` layer, so it should enclose this layer. It does not because since it has no parent layer, the current algorithm to sort layers considers it is a base layer.

With this PR layers are sorted like this:

```
no layer ->
  <class 'test_such_timing.Layer1'> ->
    <class 'test_such_timing.A system with a fast initial setup layer:layer'> ->
     test 0000: should not have obj initialized (test_such_timing.A system with a fast initial setup layer)
      <class 'test_such_timing.Layer2'> ->
        <class 'test_such_timing.having a second slow setup layer:layer'> ->
         test 0000: should have obj initialized (test_such_timing.having a second slow setup layer)
        <- <class 'test_such_timing.having a second slow setup layer:layer'>
      <- <class 'test_such_timing.Layer2'>
    <- <class 'test_such_timing.A system with a fast initial setup layer:layer'>
  <- <class 'test_such_timing.Layer1'>
<- no layer
```

One unit-test still fails, but this is because it basically verifies the problematic behavior.

There is still work to do (adding tests, splitting this into several smaller commits, etc), but if anyone has feedback about this, I'd like to hear. @jpellerin I'd be interested in your opinion, since you seem to be the main architect of this layers feature (btw, thanks for all the super detailled unit and functional tests).

### DIFF
--- a/nose2/exceptions.py
+++ b/nose2/exceptions.py
@@ -8,3 +8,8 @@ __unittest = True
 class TestNotFoundError(Exception):
 
     """Raised when a named test cannot be found"""
+
+
+class LoadTestsFailure(Exception):
+
+    """Raised when a test cannot be loaded"""
--- a/nose2/plugins/layers.py
+++ b/nose2/plugins/layers.py
@@ -3,9 +3,9 @@ import re
 
 import six
 
-from nose2 import events, util
+from nose2 import events, util, exceptions
 from nose2.suite import LayerSuite
-from nose2.compat import unittest, OrderedDict
+from nose2.compat import OrderedDict
 
 BRIGHT = r'\033[1m'
 RESET = r'\033[0m'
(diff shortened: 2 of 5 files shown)


little-dude:
For the record, here is the failing unit test: https://github.com/nose-devs/nose2/pull/267/files#diff-63735bdc208b91690ebea2cac1ae65ecR223

PR #268, opened 2016-02-16
fix #213

### LINKED ISSUE #213: such and normal test cases: mixed up call order of it.has_setup and setUpClass
As it is a little bit complicated to explain, I have created a working example as a gist: https://gist.github.com/jrast/109f70f9b4c52bab4252

I have several "normal" UnitTestCases and a few such tests. The problem is that the setUpClass calls are mixed up with the setup calls in such tests, as can be seen in the log file. The documentation states that I can not use Layer and setUpClass functions in the same TestCase, but it does not state that I cant use them side by side wich is the case in my example. So the question is: is this a bug or is there a missing hint in the documentation.

Either way I think this is a dangerous behaviour because test fixtures made in a setup call can be modified from outside of the current test case wich can lead to very unpredictable behaviour and wrong test results. One example of wrong behaviour is shown with the UniqueResource class wich is locked in the setup functions and unlocked in the teardown functions. As soon as the calls get mixed up a exception is thrown because the resource is already locked.

I realy hope this is not just a missing hint/warning in the docs, otherwise I have to rethink about how to structure my tests...

little-dude:
attempts to fix https://github.com/nose-devs/nose2/issues/213

Here is a simplified version of the `TestSuite.run` in `unittest`:
```python
    def run(self, result, debug=False):
        topLevel = False
        if getattr(result, '_testRunEntered', False) is False:
            # result._testRunEntered set to `True` when we start running the suite
            result._testRunEntered = topLevel = True

        # we iterate over the tests with the following logic:
        #      - for each test
        #          - check if it a suite:
        #               - yes => run the suite
        #               - no => check if it belongs to the same class than the previous class:
        #                     - yes => run the test
        #                     - no => run tearDown of previous class, run setup for current class, run the test
        for index, test in enumerate(self):

            if _isnotsuite(test):
                self._tearDownPreviousClass(test, result)
                self._handleModuleFixture(test, result)
                self._handleClassSetUp(test, result)
                result._previousTestClass = test.__class__

                if (getattr(test.__class__, '_classSetupFailed', False) or
                    getattr(result, '_moduleSetUpFailed', False)):
                    continue

            test(result)

        if topLevel:
            self._tearDownPreviousClass(None, result)
            self._handleModuleTearDown(result)
        return result
```

in the example given in the issue the suite looks like this:
```
<unittest2.suite.TestSuite tests=[
                <test_normal.NormalTest testMethod=test>,
                <test_normal.NormalTestTwo testMethod=test>,
                <nose2.suite.LayerSuite tests=[
                             <test_such.A system with setup testMethod=test 0000: should do something>
                ]>
]>
```

So what happens is that at the 3rd iteration, `test` is `<nose2.suite.LayerSuite tests=[<test_such.A system with setup testMethod=test 0000: should do something>]>`, which is a suite. Hence, the teardown is _not_ run.

The solution is to add logic in `nose2.suite.LayerSuite` to handle the previous test teardown, if the test class is not part of a layer. To determine if a test class is part of a layer or not, we check for a `layer` attribute, and if its value is an instance of a `LayerSuite`. The thing is I'm not sure the value will always be an instance of `LayerSuite`, so I need to make more tests.

### DIFF
--- a/nose2/suite.py
+++ b/nose2/suite.py
@@ -22,6 +22,7 @@ class LayerSuite(unittest.BaseTestSuite):
         self.wasSetup = False
 
     def run(self, result):
+        self.handle_previous_test_teardown(result)
         if not self._safeMethodCall(self.setUp, result):
             return
         try:
@@ -36,6 +37,21 @@ class LayerSuite(unittest.BaseTestSuite):
         finally:
             if self.wasSetup:
                 self._safeMethodCall(self.tearDown, result)
+
+    def handle_previous_test_teardown(self, result):
+        try:
+            prev = result._previousTestClass
+        except AttributeError:
+            return
+        layer_attr = getattr(prev, 'layer', None)
+        if isinstance(layer_attr, LayerSuite):
+            return
+        try:
+            suite_obj = unittest.suite.TestSuite()
+            suite_obj._tearDownPreviousClass(None, result)
+            suite_obj._handleModuleTearDown(result)
+        finally:
+            delattr(result, '_previousTestClass')
 
     def setUp(self):
         if self.layer is None:
--- a/nose2/tests/functional/support/scenario/layers_and_non_layers/common.py
+++ b/nose2/tests/functional/support/scenario/layers_and_non_layers/common.py
@@ -1,0 +1,60 @@
+import unittest
+import logging
+log = logging.getLogger(__name__)
+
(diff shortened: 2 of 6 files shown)


little-dude:
It's still failing when using the `uses` decorator. Added a test for this.

PR #271, opened 2016-02-21
add python 3.5 support

little-dude:


little-dude:
most of the failures are due to the `unittest.TestCase.__str__()` output begin different. The classname is now a fully qualified name as described in [PEP-3155](https://www.python.org/dev/peps/pep-3155/).

PR #282, opened 2016-03-01
test attributes -E option: make builtins available

abetkin:
Makes querying by test attributes easier:

``` python
def test(self):
    pass
test.flags = 'red', 'green', 'blue'
```

The above test can be executed with

```
python -m nose2 -E 'set(flags) >= {"green", "blue"}'
```

### DIFF
--- a/nose2/plugins/attrib.py
+++ b/nose2/plugins/attrib.py
@@ -1,3 +1,4 @@
+import sys
 import logging
 from unittest import TestSuite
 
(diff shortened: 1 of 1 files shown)


little-dude:
@abetkin thanks, I like the idea!
Would you mind mentionning it the doc ?
A functional test would be nice as well.
No pressure, I can add one or two when merging if you prefer.

abetkin:
Glad you like it. I think there might be an issue with that since builtins also will be visible when launching through `-A` option. It would be better to make builtins available only for `ContextHelper`. I think I will amend the PR and add some tests.

little-dude:
Closing since there was no activity for more than a year but feel free to re-open with some tests.

Issue #284, opened 2016-03-11
multiprocess plugin hanging during test execution


jredl-va:
We've encountered an issues lately that is quite confusing. Our CI server runs tests in parallel with 4 processes. However, it has started hanging every once in awhile. Some digging has allowed us to create an example TestSuite that highlights this behavior:

```python
import unittest

class ThreeTests(unittest.TestCase):

    def test_one(self):
        pass

    def test_two(self):
        pass

    def test_three(self):
        pass
```

running the following will show nose2 hang:
```bash
nose2 three_tests --plugin=nose2.plugins.mp -N 4
```

The hypothesis is that when running the tests in bulk (10,000+ tests in many suites), that a test suite with three tests is firing at times that is not initially sending an event to the process? 

We've done some playing with where the test suite is flattened and distributed but these seems to have consequences on overall test run times: https://github.com/nose-devs/nose2/blob/master/nose2/plugins/mp.py#L75

``` python
        # send one initial task to each process
        for proc, conn in procs:
            if not flat and self.procs_initialized is True:
                break
            if not flat and self.procs_initialized is False:
                conn.send(None)
                continue
            caseid = flat.pop(0)
            conn.send(caseid)
        self.procs_initialized = True
```

Anyone have any other ideas?

jredl-va:
Tagging @ptthiem to see if you have any suggestions as you were the main contributor to this file?

little-dude:
hi @jredl-va thanks for debugging it. I probably won't be able to work on it this week end, but I'll look into it next week if @ptthiem does not chime in in the meantime :)

ptthiem:
On windows? Or Posix?

jredl-va:
Replicated on Mac osx latest release. CI server is ubunto 14 something.

ptthiem:
Thx, jredl-va.  At least we don't have windows to deal with :D.  

Yeah the workaround would shutdown any worker processes on start up if there are not enough tests to give at least one to each out initially.  I'm not sure why would impact run-times much because a process pool is constructed for each execution of _runmp, are not persisted, and _runmp hooks into startTestRun.  Wouldn't this get run only once per execution anyway (someone)?  I am likely just missing something, trying to remember.  In any case, the loop condition is rdrs or flat.  Looks like we continue while there are tests *or* there are alive processes.  Processes appear to removed only when they return None.  Which if they never get a test and don't fail in init, then they are happily blocking on conn.recv().  That is probably why it is blocking.

[...]

PR #286, opened 2016-03-12
Branch/mp hang fix

ptthiem:
This is probably about the simplest fix for hanging when there are not enough tests for each process as demonstrated by #284 

I ran this on a directory with 10,000 copies of the test and didn't see any difference in execution time.


little-dude:
thanks a lot for this :+1:  I just squashed it to make it convenient to backport to the 0.6 branch: https://github.com/nose-devs/nose2/commit/be999d412c8dce10e255e5c4e6b49cd3840129ff

PR #288, opened 2016-03-23
Prevent crashing from UnicodeDecodeError

jmoldow:
On Python 2, `sys.stdout` and `print` can normally handle any combination of `str` and `unicode` objects. However, `StringIO.StringIO` can only safely handle one or the other.  If the program writes both a non-ASCII `unicode` string, and a non-ASCII `str` string, then the `getvalue()` method will fail with `UnicodeDecodeError` [1].

If `nose2.plugins.buffer` (which captures writes to `sys.stdout` in a `StringIO.StringIO`) is being used, that combination of writes, followed by a test failure, causes the script to suddenly abort, with the cryptic `UnicodeDecodeError`.

This fix catches `UnicodeError` when trying to get the captured output, and will replace the captured output with a warning message.

This bug is equivalent to nose-devs/nose#816 [2].

[1] <https://github.com/python/cpython/blob/2.7/Lib/StringIO.py#L258>
[2] <https://github.com/nose-devs/nose/issues/816>

### DIFF
--- a/nose2/plugins/buffer.py
+++ b/nose2/plugins/buffer.py
@@ -15,6 +15,7 @@ talk to the user.
 """
 
 import sys
+import traceback
 
 from six import StringIO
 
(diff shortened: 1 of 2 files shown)


little-dude:
@jmoldow thank you for the detailled explanation. Is this something you'd like to see in a 0.6 release or can it wait for 0.7?

jmoldow:
Doesn't matter to me.

PR #319, opened 2016-12-15
Add config section cacheing on first access, plus test of same

benj-pml:
The session object was parsing the config file anew every time the config information was accessed. There was a comment saying "FIXME" and I wanted to do something with a plugin extension that involved dynamically changing the config contents (which I couldn't do if they weren't actually stored between accesses), so I wrote this fix to cache the config section retrieved after each config file access.

It's a pretty simple change and someone presumably looked at this previously (hence the FIXME), so there may be a reason it wasn't done before! I can't see anything obvious though and the tests pass. 

Caveat: tested under python 2.7, I don't have tox set up because I did something a little odd with my nose2 installation. It's a straightforward change though and 2to3 doesn't spot anything, so I'm confident that if the tests pass under 2.7 they'll pass under everything else too.

### DIFF
--- a/nose2/session.py
+++ b/nose2/session.py
@@ -87,6 +87,7 @@ class Session(object):
         self.testResult = None
         self.testLoader = None
         self.logLevel = logging.WARN
+        self.configCache = dict()
 
     def get(self, section):
         """Get a config section.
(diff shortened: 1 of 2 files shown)


benj-pml:
I'm a little confused, the coverage check fails because the coverage on functions.py and generators.py has gone down slightly, but I haven't touched those files or changed any pre-existing tests?

katrinabrock:
Rebased, edited for clarity, and merged. Thanks!

PR #326, opened 2017-02-02
Added ability to predefine layers of fixtures and tests, and inject them into other Scenarios

### LINKED ISSUE #324: No easy way to define and utilize a reusable layer (with tests) in Such
I've been digging around the source code and haven't been able to find a clean way to define a labelled section of layers, sublayers, and tests that I can reuse in multiple places (although, I may have missed it). I'm currently trying to write some functional tests for a website, and at many points, I'll have to run the same exact tests, so being able to define those layers ahead of time and just drop them in wherever they need to be would be incredibly useful.

Ideally, I'd like to have something like this be spat out:

```
MainPage
  NavigationBar
    AboutUsLink
      displayed ... ok
    SubMenu
      Default
        MenuItem1Link
          not displayed ... ok
      Expanded
        MenuItem1Link
          displayed ... ok
OtherPage
  NavigationBar
    AboutUsLink
      displayed ... ok
    SubMenu
      Default
        MenuItem1Link
          not displayed ... ok
      Expanded
        MenuItem1Link
          displayed ... ok
```

where `NavigationBar` is a collection of layers, sublayers, and tests that is being reused in both `MainPage` and `OtherPage`.

I've tried a number of methods, but nothing seems to be working. Is there a way to do this? And if not, maybe this is a good opportunity to increase the flexibility of `Such`.

SalmonMode:
There's about a dozen ways I could have gone with this, so I figured this is as good a place as any to start the conversation to see what your vision is for nose2, and how we can make it a bit more flexible without compromising that vision.

First, I made some changes to `Scenario` so that you can determine what is inserted before and after the description of the layer, so that it doesn't have to be `"A ..."` every time and can just be left blank. 

Next I put in the `Layer` function, which works exactly like `such.A`, with the only difference that it allows you to dictate the text that comes before and after your description, with both of those being blank by default, so:

```python
with such.A("scenario") as it:
```

would spit out as:

```
A scenario
```

but:

```python
with such.Layer("Scenario") as it:
```

would spit out as:

```
Scenario
```

Then I modified the `Scenario.uses` method so that if you pass it another `Scenario` instance, it will inject the main `Group` of the `Scenario` instance you passed into the parent `Scenario`'s list of child `Group`s, as though you had define it right there.

For example, this allows you to do something like this:

```python
with such.Layer("My Common Layer") as MyCommonLayer:

    @MyCommonLayer.has_setup
    def setUp():
        pass

    @MyCommonLayer.should("have this test")
    def test():
        assert True

    @MyCommonLayer.should("also have this test")
    def test():
        assert True

    @MyCommonLayer.should("have this test, too")
    def test():
        assert True


with such.Layer("Layer A") as LayerA:

    @LayerA.has_setup
    def setUp():
        pass

    LayerA.uses(MyCommonLayer)


with such.Layer("Layer B") as LayerB:

    @LayerB.has_setup
    def setUp():
        pass

    LayerB.uses(MyCommonLayer)


with such.A("scenario") as it:

    @it.has_setup
    def setUp():
        pass

    it.uses(LayerA)
    it.uses(LayerB)
```

to get an output like this:

```
A scenario
  Layer A
    My Common Layer
      should have this test ... ok
      should also have this test ... ok
      should have this test, too ... ok
  Layer B
    My Common Layer
      should have this test ... ok
      should also have this test ... ok
      should have this test, too ... ok
```

Also, In order to facilitate identical usage no matter where you created a layer relative to where it was used, I had to also add in a `__setattr__` method to the `Scenario` class to more heavily utilize the `_helper` attribute. As a result, if I set an attribute to the `Scenario` instance, it will almost always be applied as an attribute to the `helper` object so all `Scenario` instances will be able to access it no matter where they get made (even if they're made in another module and then imported). A similar approach is taken for deleting attributes.

These changes make it so that test suites already utilizing the `it.using` method won't be affected.

Fixes #324

### DIFF
--- a/nose2/tests/functional/support/such/test_such.py
+++ b/nose2/tests/functional/support/such/test_such.py
@@ -13,6 +13,18 @@ class SomeLayer(object):
     def tearDown(cls):
         del it.somelayer
 
+
+#
+# Layers of test fixtures, tests, and other layers can be defined ahead of time
+# for later use in multiple places. They all also share the same attribute
+# namespace, so referencing `external.somelayer` will get you the same object
+# that `it.somelayer` references in the next Such Scenario.
+#
+with such.Layer('having a case inside the external fixture') as external:
+    @external.should('still have access to that fixture')
+    def test(case):
+        assert external.somelayer
+
 #
 # Such tests start with a declaration about the system under test
 # and will typically bind the test declaration to a variable with
@@ -142,10 +154,7 @@ with such.A('system with complex setup') as it:
             def test(case):
                 assert it.somelayer
 
-            with it.having('a case inside the external fixture'):
-                @it.should('still have access to that fixture')
-                def test(case):
-                    assert it.somelayer
+            it.uses(external)
 
 #
 # To convert the layer definitions into test cases, you have to call
--- a/nose2/tools/such.py
+++ b/nose2/tools/such.py
@@ -59,6 +59,10 @@ def Layer(description, desc_pre=None, desc_post=None):
       it.createTests(globals())
 
     """
+    if desc_pre is None:
+        desc_pre = ""
+    if desc_post is None:
+        desc_post = ""
     yield Scenario(
         description,
         desc_pre=desc_pre,
--- a/nose2/tools/such.py
+++ b/nose2/tools/such.py
@@ -257,6 +257,12 @@ class Scenario(object):
         else:
             setattr(self._helper, attr, value)
 
+    def __delattr__(self, attr):
+        if attr in self.__dict__.keys() or attr == "_group":
+            super(Scenario, self).__delattr__(attr)
+        else:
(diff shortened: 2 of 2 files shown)


katrinabrock:
@SalmonMode 
Are you still interested in having this feature merged? If so, can you add a doc? If not, I'll close w/o merging.

SalmonMode:
Actually, back when I thought this repo was dead, I ended up making my own tool to do this regardless of the testing framework that works similarly to Such ([shameless plug](https://github.com/salmonmode/contextional)), so I don't have a need for this anymore.

If you still want these changes, just let me know, and I'd be more than happy to add something to the docs for this.

katrinabrock:
Understandable.

In that case, others can use your plugin if they have similar needs.

PR #335, opened 2017-09-04
Change `getargspec` to `getfullargspec`

rshest:
`inspect.getargspec` is [deprecated since Python 3.0](https://docs.python.org/3/library/inspect.html#inspect.getargspec), and it breaks the test runner if using  decorators on test functions. 

`inspect.getfullargspec` is the recommended alternative:
> Use getfullargspec() for an updated API that is usually a drop-in replacement, but also correctly handles function annotations and keyword-only parameters.

In our case nose2 test runner was failing to run together with the [hypothesis](https://github.com/HypothesisWorks/hypothesis-python) property-based testing library.

This change fixes that, and possibly other similar problems that may occur.

Note that this change only affects Python 3+ code.

### DIFF
--- a/nose2/plugins/loader/functions.py
+++ b/nose2/plugins/loader/functions.py
@@ -54,7 +54,6 @@ to set ``paramList`` is with the :func:`nose2.tools.params` decorator.
 
 
 import sys
-import inspect
 import types
 
 from nose2 import util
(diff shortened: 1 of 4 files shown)


little-dude:
Nice, thank you!

Issue #337, opened 2017-10-04
Drop unsupported Python 2.6, 3.2 and 3.3?


hugovk:
## Reasons for dropping old ones

They're all EOL. Dropping 2.6 would allow a lot of `unittest` vs `unittest2` code to be dropped and simplified. Dropping 3.2 would simplify Unicode handling. Dropping 3.3 has fewer gains, but would still reduce the support burden somewhat.

* https://en.wikipedia.org/wiki/CPython#Version_history

### 2.6

* https://snarky.ca/stop-using-python-2-6/
* http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html
* http://www.python3statement.org
* Current pip 9 deprecates Python 2.6 support, pip 10 won't support it (https://github.com/pypa/pip/issues/3955)
* Not much PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

### 3.2

* pip 8 no longer supports Python 3.2 ([link](https://pip.pypa.io/en/stable/news/))
* Coverage 4 no longer supports Python 3.2 ([link](https://bitbucket.org/ned/coveragepy/issues/407/coverage-failing-on-python-325-using))
* Requests no longer supports Python 3.2
* Virtually no PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

### 3.3

* pip 10 will deprecate Python 3.3 support, pip 11 won't support it (https://github.com/pypa/pip/issues/3796)
* Very little PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

katrinabrock:
I'm fine dropping 2.6 and 3.2 for sure. Not totally convinced on 3.3. 

@little-dude thoughts?

hugovk:
Here's pip download stats for the last month from PyPI for nose2:
```
$ pypinfo --percent --pip nose2 pyversion
python_version percent download_count
-------------- ------- --------------
2.7            48.0%           21,921
3.5            19.2%            8,761
3.6            17.5%            7,988
3.4            10.5%            4,792
2.6            3.9%             1,769
3.3            0.7%               324
3.7            0.2%                89
3.2            0.0%                 8
```

katrinabrock:
I'm convinced. Merged. Thanks for your work on this! Sorry for the delayed responses.

PR #338, opened 2017-10-04
Drop unsupported Python 2.6

hugovk:
For https://github.com/nose-devs/nose2/issues/337.

## Reasons for dropping Python 2.6

* https://en.wikipedia.org/wiki/CPython#Version_history
* https://snarky.ca/stop-using-python-2-6/
* http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html
* http://www.python3statement.org
* Current pip 9 deprecates Python 2.6 support, pip 10 won't support it (https://github.com/pypa/pip/issues/3955)
* Not much PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796


katrinabrock:
merged #341

PR #339, opened 2017-10-10
Allow combination of MP and OutputBuffer plugins on Python 3.

TyberiusPrime:
You could not combine those two plugins in Python 3.4 (and presumably
any Python 3 version).
The error message when combining OutputBuffer and MP on Python 3.4
was internal error, 'state is not a dictionary', and 0 test were reported to have been run. 
It stems from a python bug (https://bugs.python.org/issue26695) that prevents
unpickling objects which return a falsy value from __getstate__
when using _pickle.

The MP plugin sends the Buffer object, and its nested Stream, back
from the spawned processes, when a string would do just as well.

This patch moves the stream -> string conversion to setTestOutcome,
which still happens in the subprocess.

It includes a test case exposing the issue.

### DIFF
--- a/nose2/plugins/buffer.py
+++ b/nose2/plugins/buffer.py
@@ -85,52 +85,57 @@ class OutputBufferPlugin(events.Plugin):
         """Stop buffering"""
         self._restore()
 
+    def _get_stream_unicode_save(self, stream, buffer):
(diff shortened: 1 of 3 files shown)


katrinabrock:
Thanks for working on this!

PR #348, opened 2017-10-23
Use specific version of python for requirements

### LINKED ISSUE #344: Automatically install per-python version requirements with `pip install nose2`
As far as I can tell, today `pip install nose2` only installs requirements in requirements.txt even if there are additional requirements for the version of python running in `requirements-x.y.txt`. Looks like these additional requirements are installed by-hand in travis: https://github.com/nose-devs/nose2/blob/master/.travis.yml#L14 , should be done in setup.py. We are already checking the version in setup.py, might as well use it for this. https://github.com/nose-devs/nose2/blob/master/setup.py#L55

artragis:
This aims to fix #344

For now, I'm not used to jython and pypy so I don't know how to handle them.

Moreover I'm not sure using the sys.version instead of `sys.version_info` is the right thing to do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,7 @@ python:
 - '3.4'
 - '3.5'
 install:
-- travis_retry pip install -r requirements-${TRAVIS_PYTHON_VERSION}.txt --allow-external
-  argparse
+- travis_retry pip install .
 - travis_retry pip install coveralls
 - travis_retry pip install coverage
 script:
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,8 @@
 import os
 import sys
 
+from os.path import dirname
+
 NAME = 'nose2'
 VERSION = open('nose2/_version.py').readlines()[-1].split()[-1].strip('"\'')
 PACKAGES = ['nose2', 'nose2.plugins', 'nose2.plugins.loader',
@@ -52,10 +54,28 @@ params = dict(
 )
 
 
-py_version = sys.version[:3]
+py_version = sys.version_info
 
 SCRIPT1 = 'nose2'
-SCRIPT2 = 'nose2-%s' % (py_version,)
+SCRIPT2 = 'nose2-%s.%s' % (py_version.major, py_version.minor)
+
+
+def parse_requirements(requirement_file):
+    requirements = []
+    with open(requirement_file) as file_pointer:
+        for line in file_pointer:
+            if line.strip() and not line.strip().startswith('#'):
+                requirements.append(line.strip())
+    return requirements
+
+
+def per_version_requirements(extra_requires_dict):
+    for current_file in os.listdir(dirname(__file__)):
+        if not current_file.startswith('requirements-') or 'docs' in current_file:
+            continue
+        python_version = current_file[len('requirements-'):-len('.txt')]
+        extra_requires_key = ':python_version == "{}"'.format(python_version)
+        extra_requires_dict[extra_requires_key] = parse_requirements(current_file)
 
 try:
     from setuptools import setup
@@ -63,15 +83,15 @@ except ImportError:
     from distutils.core import setup
 else:
 
-    REQS = ['six>=1.1']
-
     params['entry_points'] = {
         'console_scripts': [
             '%s = nose2:discover' % SCRIPT1,
             '%s = nose2:discover' % SCRIPT2,
         ],
     }
-    params['install_requires'] = REQS
+    params['install_requires'] = parse_requirements('requirements.txt')
     params['test_suite'] = 'unittest.collector'
+    params['extras_require']['doc'] = parse_requirements('requirements-docs.txt')
+
 
 setup(**params)
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ def parse_requirements(requirement_file):
 
 
 def per_version_requirements(extra_requires_dict):
-    for current_file in os.listdir(dirname(__file__)):
+    for current_file in os.listdir(dirname(__file__) or '.'):  # the '.' allows tox to be run locally
         if not current_file.startswith('requirements-') or 'docs' in current_file:
             continue
         python_version = current_file[len('requirements-'):-len('.txt')]
@@ -92,6 +92,6 @@ else:
     params['install_requires'] = parse_requirements('requirements.txt')
     params['test_suite'] = 'unittest.collector'
     params['extras_require']['doc'] = parse_requirements('requirements-docs.txt')
-
+    per_version_requirements(params)
 
 setup(**params)
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist=py27,py33,py34,pypy,docs,self27,cov27
 
 # Default settings for py27, py33, py34 and pypy
 [testenv]
-deps=-r{toxinidir}/requirements.txt
+deps=.
 commands=python -m unittest discover []
 
 [testenv:jython]
@@ -14,23 +14,21 @@ commands=unit2 discover []
 [testenv:docs]
 basepython=python2.7
 changedir=docs
-deps=-r{toxinidir}/requirements-2.7.txt
-     -r{toxinidir}/requirements-docs.txt
+deps= .[doc]
 commands=sphinx-build -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
 
 [testenv:py27]
-deps=-r{toxinidir}/requirements-2.7.txt
+deps={toxinidir}
 
 [testenv:self27]
 basepython=python2.7
-deps=-r{toxinidir}/requirements-2.7.txt
+deps= {toxinidir}
 setenv=PYTHONPATH={toxinidir}
 commands=python -m nose2.__main__ []
 
 [testenv:cov27]
 basepython=python2.7
-deps=coverage>=3.3
-     -r{toxinidir}/requirements-2.7.txt
+deps= {toxinidir}
 commands=coverage erase
          coverage run -m unittest discover []
          coverage report --include=*nose2* --omit=*nose2/tests*
--- a/requirements-2.7.txt
+++ b/requirements-2.7.txt
@@ -1,2 +1,1 @@
 mock
--r requirements.txt
--- a/requirements-3.2.txt
+++ b/requirements-3.2.txt
@@ -1,1 +1,0 @@
--r requirements.txt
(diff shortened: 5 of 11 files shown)


katrinabrock:
Cool, thanks. Let's update the.`.travis.yml` too to use this instead of its own thing. That will also tell us if it's working for pypy.

artragis:
I tried something that can help. But as travis does not use tox, I can't see anything change.

katrinabrock:
What you did is basically what I had in mind.

.travis.yml
```diff
@@ -11,7 +11,7 @@ python:
 - '3.4'
 - '3.5'
 install:
-- travis_retry pip install -r requirements-${TRAVIS_PYTHON_VERSION}.txt --allow-external
+- travis_retry pip install -e .
   argparse
```
katrinabrock: I think argparse is the argument for --allow-external option so should delete both or neither.

artragis: thanks a lot.

katrinabrock:
Looking at the travis logs for this pr vs others, it seems like the installation is already happening, that these statements are *just* for requirements, so maybe try deleting that `pip install` line all together?

After this, tox integration is definitely the next step. If you want to play around with travis config, feel free to open few dummy PRs to kick off builds.

artragis:
For now, I just deal with travis. Your first comment just made me realize tox was not yet in there so my changes were not tested.

The next run should pass.

katrinabrock:
Looks like it's passing :-) Any reason you're installing in dev mode? `-e`

artragis:
It's just an habit. If you want me to remove that, I can.

katrinabrock:
Yeah, now that it's passing, I'd like to make the code maybe a tiny bit more readable. Don't have time right this moment, but I'll take a look and give more specific comments w/in 24-48hrs.

.travis.yml
```diff
@@ -11,8 +11,7 @@ python:
 - '3.4'
 - '3.5'
 install:
-- travis_retry pip install -r requirements-${TRAVIS_PYTHON_VERSION}.txt --allow-external
-  argparse
+- travis_retry pip install -e .
```
katrinabrock: Yeah, should probably not test in dev mode.

setup.py
```diff
@@ -52,26 +54,48 @@
...
+
+        if py_version.major == 3:
+            version_specific_requirements_path = 'requirements-3.5.txt'
+        else:
+            version_specific_requirements_path = 'requirements-2.7.txt'
+    with open(os.path.join(directory, version_specific_requirements_path)) as requirements_file:
+        REQS.update(list(filter(lambda s: s and not s.startswith('#'), requirements_file.readlines())))
+    with open(os.path.join(directory, 'requirements.txt')) as requirements_file:
+        REQS.update(list(filter(lambda s: s and not s.startswith('#'), requirements_file.readlines())))
+    print(REQS)
```
katrinabrock: Is this a debugging message?

setup.py
```diff
@@ -52,26 +54,48 @@
...
-py_version = sys.version[:3]
+py_version = sys.version_info
+is_pypy = "PyPy" in sys.version
 
 SCRIPT1 = 'nose2'
-SCRIPT2 = 'nose2-%s' % (py_version,)
+SCRIPT2 = 'nose2-%s.%s' % (py_version.major, py_version.minor)
+
 
+def _get_specific_requirements():
```
katrinabrock: This function should probably take version as input and return a list of requirements instead using and setting variables outside of its scope.

Also would be nice to have a docstring, even a oneliner.

setup.py
```diff
@@ -52,26 +54,48 @@
...
+        if py_version.major == 3:
+            pypy_name = 'pypy3'
+        version_specific_requirements_path = 'requirements-{}.txt'.format(pypy_name)
+    if not os.path.exists(os.path.join(directory, version_specific_requirements_path)):
+
+        if py_version.major == 3:
+            version_specific_requirements_path = 'requirements-3.5.txt'
+        else:
+            version_specific_requirements_path = 'requirements-2.7.txt'
+    with open(os.path.join(directory, version_specific_requirements_path)) as requirements_file:
```
katrinabrock: Maybe use `pip.req`?
http://jelly.codes/articles/python-pip-module/

katrinabrock: Otherwise, this should at least be a function to avoid duplicate code.

setup.py
```diff
@@ -52,26 +54,48 @@
...
+is_pypy = "PyPy" in sys.version
 
 SCRIPT1 = 'nose2'
-SCRIPT2 = 'nose2-%s' % (py_version,)
+SCRIPT2 = 'nose2-%s.%s' % (py_version.major, py_version.minor)
+
 
+def _get_specific_requirements():
+    directory = dirname(__file__)
+    version_specific_requirements_path = 'requirements-{}.{}.txt'.format(py_version.major, py_version.minor)
```
katrinabrock: Is it possible to make these nested if statements into just one if...elif...elif...else ?
Or maybe something like.
```
if is_pypy:
    pypy_name = "pypy"
else:
   pypy_name = ""
   version_specific_requirements_path = 'requirements-{}{}.{}.txt'.format(pypy_name, py_version.major, py_version.minor)
```
and then rename the requirements files accordingly. 

Overall goal would be to make logic easier to decipher at first glance.

setup.py
```diff
@@ -52,26 +54,48 @@
...
     params['entry_points'] = {
         'console_scripts': [
             '%s = nose2:discover' % SCRIPT1,
             '%s = nose2:discover' % SCRIPT2,
         ],
     }
-    params['install_requires'] = REQS
+    params['install_requires'] = list(REQS)
     params['test_suite'] = 'unittest.collector'
+    params['extras_require']['doc'] = ['Sphinx>=1.0.5']
```
katrinabrock: so, [this answer](https://stackoverflow.com/questions/19559247/requirements-txt-depending-on-python-version) suggests using this value to solve this whole problem. Not sure if it will work with `pypy` but maybe worth a shot?

setup.py
artragis: because of pypy3 I think you meant : 

```python
pypy_name = ''
if is_pypy:
    pypy_name = "pypy"
version_specific_requirements_path = 'requirements-{}{}.{}.txt'.format(pypy_name, py_version.major, py_version.minor)
```

With that I have to create multiple requirements file for pypy3.3/pypy3.2/pypy3.5, are you ok with that? (pypy doc : http://doc.pypy.org/en/latest/index-of-release-notes.html#cpython-3-5-compatible-versions)

setup.py
artragis: Yes, I will deduplicate. Thanks for pip.req, I didn't know it :)

setup.py
artragis: I did not knew it. I think I will make a full rework.

setup.py
artragis: in fact `pip.req` seems not to be a good idea : https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py : 

>  You really don't want to do this. Speaking as a pip maintainer pip does not support being called as an API like this at all. In fact pip 1.6 (next version at this time) moves this function. – Donald Stufft Mar 26 '14 at 0:59

setup.py
artragis: Yes, removed, sorry.

.travis.yml
artragis: fixed.

setup.py
katrinabrock: The way you ended up doing it looks great and passes travis.

setup.py
katrinabrock: You are correct.
but :-( would be nice if they supported this

setup.py
```diff
@@ -52,26 +54,44 @@
...
+def parse_requirements(requirement_file):
+    requirements = []
+    with open(requirement_file) as file_pointer:
+        for line in file_pointer:
+            if line.strip() and not line.strip().startswith('#'):
+                requirements.append(line.strip())
+    return requirements
+
+
+def per_version_requirements(extra_requires_dict):
```
katrinabrock: Am I missing something? Is this getting called anywhere? If not, how is travis passing? <confused>

Also, you don't need to pass in the old version, you can just have something like
```
params['extra_require'].update(per_version_requirements())
```

tox.ini
```diff
@@ -3,7 +3,7 @@ envlist=py27,py33,py34,pypy,docs,self27,cov27
 
 # Default settings for py27, py33, py34 and pypy
 [testenv]
-deps=-r{toxinidir}/requirements.txt
+deps=-e {toxinidir}
```
katrinabrock: `-e` still here...have you tested it with tox? I haven't been able to make tox generally happy so trusting the committer until we get tox integrated with travis.

setup.py
artragis: you don't miss anything.

tox.ini
artragis: For me tox run in a quite verbose way and is just failing in the doc env.
I commit something that makes it working better but tox is not totally ready I think.

setup.py
artragis: fixed

katrinabrock: Weird that it was passing though.

tox.ini
katrinabrock: Is it closer to passing than on master, the same, or worse for you?  If this makes it better, I'll merge. Not going to release until I get tox passing on travis though.

artragis: we've git exactky the same error in fact. 
I also think that one problem we have for documentation is that we use a very old version of sphinx + we do not add the "sphinx_rtd_theme" module.

artragis: I think I found what I missed.

katrinabrock:
Current state looks good to me.
Btw - are you aware that you're committing under a different name?
image

setup.py
artragis: In fact, tox is not used so the build is passing but tox is out of service.
I commit once the full tox run is over.

artragis:
Hello, for the name : yes : we use nose2 at work so I use my "working" alias instead of the pseudo I used to (artragis) If you want me to amend, I can.

katrinabrock:
I don't care one way or the other.

setup.py
katrinabrock: yes, but your travis should still be running this code when you do `pip install .` Then the tests should fail if proper requirements are not installed.

(Wouldn't fail with the old `.travis.yaml`, this should have failed with the new one)

artragis:
Ok I come back with a fix to tox as soon as possible :)

PR #349, opened 2017-10-24
Replace cov-core with coverage plugin

### LINKED ISSUE #336: Switch away from cov-core; maybe use pytest-cov
`cov-core` has been a dead project for 3 years. Although it doesn't allow for new issues to be filed, there's [at least one problem with its handling of `.coveragerc`](https://stackoverflow.com/questions/41956688)

I would be interested in checking whether or not `pytest-cov` works, since it presumably has most, if not all, of what `cov-core` used to provide.

sirosen:
`cov-core` has been unmaintained for 3 years, but the [`coverage` project](https://pypi.python.org/pypi/coverage) is getting active work as of 2017.
Try using `coverage` instead of `cov-core`. (This seems like a better solution than `pytest-cov`, mentioned in #336, which is a `pytest` plugin).
Closes #336

Notes:
- `coverage` supports all of the reporting modes of `cov-core`, but how do we verify that they all work the same in all cases...?
- Still allows "term-missing" to get "show_missing" behavior, to be backwards compatible, but otherwise will rely on coveragerc specification of "show_missing"
- `coverage` matches the plugin module name, so for py2 to work it needs a future import of `absolute_import`
- `coverage` doesn't accept reporting modes as an argument, but requires explicit calls to its reporting methods, so imitate the same dispatch that we got under `cov-core`
- `cov-core` was running between plugin initialization and `createdTestSuite` getting called. It's not obvious from docs whether or not `coverage` supports multiple start/stop calls before reporting, so I've removed this. Was having coverage measurement during this period important?


katrinabrock:
> how do we verify that they all work the same in all cases...?
If it gets roughly the same answer, should be fine.

katrinabrock:
> coverage matches the plugin module name, so for py2 to work it needs a future import of absolute_import

This is fine. 

Do we know why coveralls was triggered so many times? Did you do a bunch of fixups or something?

sirosen:
> Do we know why coveralls was triggered so many times? Did you do a bunch of fixups or something?

Nope. This is pretty weird looking -- could Travis trigger it multiple times maybe?

I was so excited to see a bunch of comments on my PR, now so crestfallen. 😞

katrinabrock:
maybe, looks like there is a retry mechanism in travis:
https://github.com/nose-devs/nose2/blob/master/.travis.yml#L15

So my guess would be that travis thinks coveralls is not working so keeps retriggering. But in that case surprised that travis passing.

Maybe try messing around with travis config and see if you can get rid of this behavior?

sirosen:
The retry seems to apply to the install of `coveralls` though?
It looks to me like coveralls gets invoked once [after each successful build](https://github.com/nose-devs/nose2/blob/f9851a2841b21b4abe98d52b0928700b28aaf937/.travis.yml#L19-L20)

Job output is consistent with that, showing things like:
```bash
$ coveralls
Submitting coverage to coveralls.io...
Coverage submitted!
Job #536.1
https://coveralls.io/jobs/30534515
```

only once, but for each job in the build.

Note: the Travis build matrix has 6 parts, but I see only 5 comments here, so it's not triggering the comment on each job in the build. Something else is messed up.

---

I'm reluctant to mess around with Travis config because it _looks right_, and I think this is a service-side bug, probably with the Coveralls.
That's consistent with [this ancient and unresolved issue](https://github.com/lemurheavy/coveralls-public/issues/491)

Also, I'd rather not block this work (which is unrelated) on Travis/Coveralls mayhem.

katrinabrock:
Not getting this coveralls behavior on other PRs.  I did just merge a commit that changed the travis config. :-/

Can't tell from coveralls issue if it's intermittent or if there's a workaround.

If you have time, can you try the following:
1) make another branch with the exact same changes as this branch and make a PR
2) make another branch with no meaningful changes to master and make a PR

See if we can reproduce this issue.

sirosen:
I realized that I had a change that I staged but never committed. Out of curiosity, I've pushed an amended commit.

I'll also open other PRs to see if they behave that way.

katrinabrock:
(for posterity)
Re: Multiple coveralls comments
Looks like this was a transient coveralls issue unrelated to these changes.

Re: Decrease in coverage
Good! If coverage increases substantially, I am concerned.

PR #350, opened 2017-10-24
make tox run well and use travis-tox

artragis:
there was a typo on setup.py extras_require that was easyly fixed
then I had to change requirements_docs as rtd_theme is no longer ebedded in Sphinx

Here is the result I get for now : 

```
  py27: commands succeeded
SKIPPED:  py33: InterpreterNotFound: python3.3
SKIPPED:  py34: InterpreterNotFound: python3.4
  pypy: commands succeeded
  docs: commands succeeded
  self27: commands succeeded
  cov27: commands succeeded
  congratulations :)
```

I put this PR in WIP because I still have few questions:

- why don't we have a py35 (and even a py36 !) env?
- why do we have a `cov27` env instead of just running the coverage on `py{x}`
- why do we have the `self27` (and not `self{other_envs}`) as we do not have it on the former travis.yml?


katrinabrock:
> why don't we have a py35 (and even a py36 !) env?

Because this project has been pretty much unmaintained for a long time.

>why do we have a cov27 env instead of just running the coverage on py{x}
why do we have the self27 (and not self{other_envs}) as we do not have it on the former travis.yml?

No clue. @little-dude or @jpellerin might know. See what you can learn from commit messages.

artragis:
Thanks, I'll update that tomorrow

katrinabrock:
@artragis you should join the gitter channel https://gitter.im/nose2/

setup.py
```diff
@@ -92,6 +92,6 @@ def per_version_requirements(extra_requires_dict):
     params['install_requires'] = parse_requirements('requirements.txt')
     params['test_suite'] = 'unittest.collector'
     params['extras_require']['doc'] = parse_requirements('requirements-docs.txt')
-    per_version_requirements(params)
+    per_version_requirements(params['extras_require'])
```
katrinabrock: I still think something is not quite right here. Looks like you're not doing anything with the return value.

artragis: There is no return value : I modify the dictionnary in place. Do you prefere I make something like `params['extras_require'].update(per_version_requirements)`?

katrinabrock: Yes, more explicit.

katrinabrock: Or I guess rename to `add_per_version_requirements`

artragis: done.

katrinabrock: Haha *or* as in modifying in place is adding requirements to an existing dictionary. A function called `per_version_requiments` I would expect to return something. This is really nitpicking though, you can leave as is.

artragis: ok

PR #356, opened 2017-11-01
Improve install docs for coverage-plugin

### LINKED ISSUE #355: Provide more useful message about missing `coverage-plugin`
When I invoke `nose2 --with-coverage` when `nose2` and `coverage` are installed, I get this unhelpful error message:

```
$ nose2 --with-coverage
Warning: you need to install [coverage-plugin] extra requirements to use this plugin
```
This message would be more helpful if it included instruction son how to install `coverage-plugin`, or what parameters to load it.

The [Test coverage reporting](http://nose2.readthedocs.io/en/latest/plugins/coverage.html) section of the nose2 documentation doesn't mention the need to install `coverage-plugin`, or how to install it, so it looks like the documentation is incorrect.

That documentation also implies that `coverage-plugin` is loaded by default, given that it appears in the [Built-in and loaded by default](http://nose2.readthedocs.io/en/latest/plugins.html#built-in-and-loaded-by-default) section.

sirosen:
The coverage plugin docstrings should document the use of the `nose2[coverage-plugin]` extra, not direct installation of `coverage`. This way, the docs work for all versions.
Additionally, improve the ImportError warning message to properly explain installation of `nose2[coverage-plugin]`.

Because this is a common point of confusion with extras, the example in the module docstring notes how to specify the version of `nose2` at the same time as the extra.

Resolves #355

### DIFF
--- a/nose2/plugins/coverage.py
+++ b/nose2/plugins/coverage.py
@@ -1,11 +1,11 @@
 """
 Use this plugin to activate coverage report.
 
-To use this plugin, you need to install ``coverage``:
+To use this plugin, you need to install ``nose2[coverage-plugin]``. e.g.
 
 ::
 
-    $ pip install coverage
+    $ pip install nose2[coverage-plugin]>=0.6.5
 
 
 Then, you can enable coverage reporting with :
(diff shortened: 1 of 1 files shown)


katrinabrock:
Awesome, thanks!

PR #359, opened 2017-11-03
Fix coverage reporting error: deferred too long

sirosen:
Waiting until `startTestRun` to start collecting coverage reporting information is wrong! Upon closer inspection, it becomes obvious that if coverage collection waits until then, it will miss any and all import-time activity which happens while loading the testsuite. Since a lot of tests are built off of real application code, that's a lot.

Adding this fix right away, and will try to write a testcase for it later. Confirmed against another project's testsuite that this gets *exactly* the same output as cov-core did.

### DIFF
--- a/nose2/plugins/coverage.py
+++ b/nose2/plugins/coverage.py
@@ -81,11 +81,9 @@ class Coverage(Plugin):
 
         self.covController = coverage.Coverage(source=self.covSource,
                                                config_file=self.covConfig)
-
-    def startTestRun(self, event):
-        """Resume coverage collection before running tests."""
-        if self.covController:
-            self.covController.start()
+        # start immediately (don't wait until startTestRun) so that coverage
+        # will pick up on things which happen at import time
+        self.covController.start()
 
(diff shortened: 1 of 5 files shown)


sirosen:
My attempts to write a testcase have been stymied. The behavior inside of the nose2 testsuite doesn't match what I see when I actually run it, and I don't understand why.
The failing test is easy to write, it's just two files:

```
# pkg/__init__.py
foo = 1
```
and
```python
from pkg import foo

def test():
    assert foo == 1
```

When you _actually run_ (as in `nose2 --with-coverage --coverage=pkg`), the coverage result comes out right.
So I can validate the fix... BUT if I do this in the testsuite, the imports aren't measured.
I think `pkg` gets correctly loaded after `nose2.plugins.coverage.createTests` runs, based on some logging, which killed my only working theory.

sirosen:
Added a failing test but wrapped it in a `@skip` for now and wrote a FIXME comment.
It includes a marginal addition of logging from attempts to track this down.

Issue #361, opened 2017-11-05
Update the README/Community Discussion/Existential Crisis


katrinabrock:
I don't think we can really say that folks reading the readme these days are

> witnesses at the new birth of nose, mark 2

Most of the other info there is also false.

I am hereby opening the discussion on where are we and where are we going?

Specifically:
* Would we actually recommend people use nose2 over pytest for new projects?
* If so why?
* If not, what is the role of this project?
* Do we want to stabilize to a 1.0.0 at some point? If so, what needs to happen before we get there?

sirosen:
> - Would we actually recommend people use nose2 over pytest for new projects?
> - If so why?

Doing so would demand some rationale. We have to be honest with ourselves about the facts that
- there's nothing that `nose2` can do that `pytest` cannot do
- the development community is a lot larger and more active around `pytest` and its plugins.

`Such` was a cool idea, and I have a project making heavy use of Layers to get more layers of setup/teardown than I could with unittest, but it's hard to make a case for using those tools over [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html).

> - If not, what is the role of this project?

*shrugs*
Bugfixes and maintenance for those still using it, and accepting new features if anyone wants to add value to projects already using `nose2`.

> - Do we want to stabilize to a 1.0.0 at some point? If so, what needs to happen before we get there?

The traditional [semver advice would suggest that nose2 should release 1.0](http://semver.org/#how-do-i-know-when-to-release-100).
Given the amount of discussion and testing in Gitter prior to releasing `0.7.0`, it seems like the mentality of not wanting to break people's builds is already there.

katrinabrock:
I agree with @sirosen response to nose2 vs pytest. 

Follow up question: what about nose2 vs nose?

Seems like there are enough of us where our companies are basically locked in to nose2 that there is capitalist interest in keeping it minimally maintained even if it is not competitive with pytest. Is the same true for nose? If not, that alone is an advantage and moving from nose to nose2 is going to be much easier than moving from nose to pytest. We can make an effort to make it even easier than it is today.

sirosen:
> moving from nose to nose2 is going to be much easier than moving from nose to pytest

I'm not sure that this is true today, but it sounds like you're saying we could make it a priority to make it easier to go `nose -> nose2` than `nose -> pytest`? It's a way to differentiate the project, at least.

[...]

PR #365, opened 2017-11-09
Do maintanence on developer setup and readme

### LINKED ISSUE #358: Get rid of bootstrap.sh (?); maybe add Makefile
Working on #357 made me take notice of this script. As far as I can tell, it's a script which invokes... `python setup.py develop`
The rest of it is window dressing. There's no need to call `pip` explicitly since setuptools handles the dependencies, it doesn't actually install sphinx dependencies, and forcing the use of virtualenvwrapper is unnecessary (plain virtualenv works just fine).

I think it should be removed.

---

If there is to be any workflow automation, I would advocate for a makefile with the following targets:
- `make help`: echo each target with description
- `make develop`: create virtualenv, `setup.py develop`, install `docs-requirements.txt`, install `tox`
- `make docs`: submake to run sphinx
- `make test`: invoke `tox`

None of this is necessary, of course, but it beats the pants off of a script that does only one part of this and does it poorly.

This big selling point for this is that someone new to the project can run `make test`, and have that install everything and run `tox` for them.

sirosen:
Resolve #358 and #361 

- Remove `bootstrap.sh`, `bootstrap.bat`, `test.sh`
- Add a Makefile where `make test` is a veneer over `virtualenv ...; pip install tox; tox`
- Update the README to reflect the weak consensus established in #361 

I think the removals are good cleanup, the `Makefile` is a suggestion (so feel free to reject it; I'd just squash it out of existence), and we need to do _something_ with the README, though it's not clear that this is the right thing.


katrinabrock:
README LGTM

I want to actually test the make file. (Testing it in travis would be ideal.)

Makefile
```diff
@@ -0,0 +1,27 @@
+# control the py version used to install tox
+# use like
+#   NOSE2_PYTHON_VERSION=python3.6 make docs
+NOSE2_PYTHON_VERSION?=python
+VIRTUALENV=.venv-$(NOSE2_PYTHON_VERSION)
```
katrinabrock: Add these folders to gitignore.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+help:
+	@echo 'Easy setup for local nose2 development'
+	@echo 'make help    Show this help'
+	@echo 'make docs    Build docs'
+	@echo 'make test    Test nose2'
+	@echo 'make clean   Cleanup'
+
+$(VIRTUALENV)/bin/tox:
+	virtualenv --python "$(NOSE2_PYTHON_VERSION)" $(VIRTUALENV)
```
katrinabrock: What happens if virtualenv is not installed?

README.rst
```diff
@@ -27,24 +27,51 @@ Welcome to nose2
...
-plugins branch, and overrides various unittest2 objects.
+Workflow
+--------
 
-You are witnesses at the new birth of nose, mark 2. Hope you enjoy our
-new direction!
+If you want to make contributions, you can use the ``Makefile`` to get started
+quickly and easily::
+
+    # no prior setup necessary!
```
katrinabrock: Well.... You need to have a supported version of python installed. Also maybe add a note here that (maybe obviously) tox can only test locally against versions of python you have installed, but we'll only merge if tests pass for all supported versions.

katrinabrock:
Merged with #368

PR #366, opened 2017-11-09
Update setuptools invocation to handle coverage

### LINKED ISSUE #317: Coverage plugin does not work with "setup.py test"
I am trying to use "python setup.py test" to run unit tests and get a coverage report. I have set the `test_suite` argument to the `setup` method to `nose2.collector.collector`, and configured a nose2.cfg file with the following:

`[coverage]`
`always-on = True`
`coverage-report = xml`

When I run "nose2" from the command line, it runs the coverage and outputs the coverage.xml file, but when I run it from setup.py, it ignores the coverage section. I can tell it's responding to the nose2.cfg file because if I add junit-xml configuration, it works using both methods, but coverage only seems to work via the nose2 command line.

sirosen:
Setuptools invocation runs differently from command-line usage. This is, itself, a potential source of grief.
Try to make it behave more like `nose2` invocation by doing the following:
- add ~/.{unittest,nose2}.cfg to loaded config
- trigger some events nose2 plugins want to use
- fix nose2.plugins.coverage to handle those events without handleArgs

Unfortunately, it looks like unifying these two modes of use more would be a significant effort, and it's not obvious that the payoff is worth the reward.

It is really easily possible that this change could break a 3rd party plugin under setuptools, since it skips the handleArgs event, but triggers some new events. However, 3rd party plugins could already be broken under setuptools by getting loaded and then not seeing these events...

Other than `nose2.plugins.coverage`, the only plugin I see using `handleArgs` is `nose2.plugins.attrib`. But attributes can't be selected if we're not processing CLI args anyway, so this is a non-issue.

Resolves #317

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+* Running `nose2` via `setuptools` will now trigger `CreateTestsEvent` and `CreatedTestSuiteEvent`
+
 0.7.3
 -----
 * Fixed
--- a/nose2/collector.py
+++ b/nose2/collector.py
@@ -1,12 +1,19 @@
+import os
 import sys
 import unittest
 
-from nose2 import loader, runner, session
+from nose2 import loader, runner, session, events
 from nose2.main import PluggableTestProgram
 __unittest = True
 
 
 def collector():
+    """
+    This is the entry point used by setuptools, as in::
+
+      python setup.py test
+
+    """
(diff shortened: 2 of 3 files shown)


katrinabrock:
Can you document how this may potentially break 3rd party plug-ins in the changelog?

sirosen:
Rebased and added requested info to the changelog.

katrinabrock:
Seems reasonable. CI is now broken though. :-/

katrinabrock:
commits merged in #389

Issue #109, opened 2013-07-19
the multiprocess plugin fails to give a useful error for unimportable test modules


sanga:
So if a test module is not importable you wind up with the following output:

```
test_not_found (nose2.loader.LoadTestsFailure) ... ERROR
test_not_found (nose2.loader.LoadTestsFailure) ... ERROR

======================================================================
ERROR: test_not_found (nose2.loader.LoadTestsFailure)
----------------------------------------------------------------------
AttributeError: 'module' object has no attribute 'ModuleImportFailure'
```

The reason for this so far as I can tell is that the mp plugin first imports the tests in the main process and then distributes those tests across the child processes by name, where they are again imported. The problem arises because for a module which can't be imported, nose2 creates a testsuite with a method that will just raise the import exception when run. Like so:

```
unittest2.suite.TestSuite tests=[<nose2.loader.ModuleImportFailure testMethod=v2.test_token>]
```

 mp.MultiProcess._flatten ends up making those into the following list of 'test names'

```
flat: ['nose2.loader.ModuleImportFailure.v2.test_object_relations',
'nose2.loader.ModuleImportFailure.v2.test_devices...
```

and when the child process tries to import the name 
nose2.loader.ModuleImportFailure.v2.test_object_relations
it fails as that test doesn't exist.

And so the root cause of the import error is lost.

so how might one fix this?
in a way it doesn't make sense to farm those tests that can't be imported
out to the child nose processes, as we already know that they fail, so we
could filter them out already before the _flatten process
my attempts to do that though haven't so far succeeded

It might be easier to just change the test name in that case (util.test_id I think does that) so that nose2.loader.ModuleImportFailure would be removed so that the child process could actually find the test to import, and so that would catch the import error. Though test names must match in both the main and the child processes.

moosingin3space:
Has any progress been made on this?

nicolasdomeniconi:
Looks like this is still an issue.

cdman:
This still happens with nose2 0.6.4

jaredj:
Still in 0.6.5 as well

ibnIrshad:
At least a quick google search of the error leads to this issue page, which helps in diagnosing the problem.

TyberiusPrime:
I have a pretty ugly hack to get the right output (see below, it does pass the test suite...), 
but I think the methodology is sound.

```patch
--- a/nose2/plugins/mp.py	Tue Oct 10 10:20:56 2017 +0200
[...]

PR #367, opened 2017-11-09
Variation of Fix for 109 and Testcase

ptthiem:
This is the simplest addition of the proposed fix in #109 .  The MP code needs some love.


katrinabrock:
Changes look reasonable and readable to me. Based on the commit message, I'm guessing it doesn't meet you expectations yet. Ping me when it does and I'll take another look before merging.

ptthiem:
That PR is probably fine and the test-case will be reusable.   I went down a rabbit hole and am mulling over a few things.

PR #369, opened 2017-11-11
[mp] Clean some of the TODO comments and Fixed #167

### LINKED ISSUE #167: MP plugin does't call startSubprocess in subprocess
My very quick diagnosis is:

In [mp.py](https://github.com/nose-devs/nose2/blob/master/nose2/plugins/mp.py):
- The new hook methods are registered on [line 27](https://github.com/nose-devs/nose2/blob/master/nose2/plugins/mp.py#L27)
- However that is only called in the master process
- Hence, when in the subprocess on [line 197](https://github.com/nose-devs/nose2/blob/master/nose2/plugins/mp.py#L197) we register the plugins, the session doesn't know to look for `startSubprocess` (or `stopSubprocess` for that matter)
- Hence when it tries to dispatch it on [line 202](https://github.com/nose-devs/nose2/blob/master/nose2/plugins/mp.py#L202) it won't call the plugins

ptthiem:
Fixed #167

### DIFF
--- a/nose2/plugins/mp.py
+++ b/nose2/plugins/mp.py
@@ -80,8 +80,8 @@ class MultiProcess(events.Plugin):
             if testid.startswith(failed_import_id):
                 self.cases[testid].run(result_)
 
-        # XXX The length of the filtered list needs to be known
-        # for _startProcs, until this can be cleaned up.  This
+        # XXX Process-Handling: The length of the filtered list needs to be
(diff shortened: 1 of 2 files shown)


ptthiem:
There is a small warning being thrown by the test, but it seems to be working.  I'll check it tomorrow afternoon

katrinabrock:
Somehow I missed when this was reopened. LGTM

PR #380, opened 2017-12-28
Respect fail_under coverage config

### LINKED ISSUE #285: coverage: fail_under and non-zero exit code
I may be doing something wrong, but it looks like the `fail_under` param in coverage is not being respected by nose2. I have this set to 100, but still get a zero return code from nose2.

For CI, etc., it would be helpful to see a non-zero return code for inadequate coverage.

sirosen:
Resolves #285

- Nose2 ResultSuccessEvent events can only indicate exit(0) or exit(1),
  so without a brand new exit status handling mechanism, we can't get
  exit(2) which `coverage` does for fail_under. *This is fine*.
  Unless we expect nose2 to bend over backwards to reproduce all of the
  behaviors of coverage, this just needs to be documented.
- Previously, coverage was computed during afterSummaryReport. However,
  it must be computed before wasSuccessful in order to produce the
  correct resulting status with fail_under. Therefore, coverage is now
  computed during beforeSummaryReport, its output is buffered in a
  StringIO, and the result (for text output) is printed during
  afterSummaryReport. This should have no externally visible behavioral
  change.
- Add support to coverage tests for checking exit status, so that we can
  assert exit(1) on the fail_under test. Add a relevant test using this
  capability.
- Add documentation to coverage plugin detailing exit(2) vs exit(1)
  behavior, and leaving room for more `coverage` vs `nose2`
  discrepancies.

### DIFF
--- a/docs/plugins/coverage.rst
+++ b/docs/plugins/coverage.rst
@@ -3,3 +3,18 @@ Test coverage reporting
 =======================
 
 .. autoplugin :: nose2.plugins.coverage.Coverage
+
+Differences From coverage
+-------------------------
+
+The ``coverage`` tool is the basis for nose2's coverage reporting. nose2 will
+seek to emulate ``coverage`` behavior whenever possible, but there are known
+cases where this is not feasible.
+
+If you need the exact behaviors of ``coverage``, consider having ``coverage``
+invoke ``nose2``.
+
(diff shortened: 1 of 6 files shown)


katrinabrock:
commit merged in #389

PR #387, opened 2018-02-07
Fixes reproducibility issues in MP plugin

### LINKED ISSUE #381: [Debian] Doc build depends on # of core on build machine (mp)
Hey,

I'd like to report two reproducibility issues I met with nose2.

MAINTAINER NOTE: Created https://github.com/nose-devs/nose2/issues/382 to track first issue. This thread should be used to track second issue.

The first is with the manpage. The order of the arguments seem to be non-deterministic as they rely on a set that is expressed in file `nose2/session.py` line 34. all_ being a set, the plugins order is inconsistent between multiple runs, and so the order of the usage from -h option is non-deterministic either. I recommend replacing all_ by a `sorted(all_)` in this line.

The second issue is with the documentation of `nose2/plugins/mp.py`. The documentation is built including the default value for processes configuration variable, which is multiprocessing.cpu_count(), and thus depending on the machine where the documentation is built. Also, this function might raise a NotImplementedError. Shouldn't the value be fixed to, let's say, 2, and eventually changed by the user?

A backward-compatible patch is maybe possible:

```
diff --git a/nose2/plugins/mp.py b/nose2/plugins/mp.py
index b712bb3..b6317b4 100644
--- a/nose2/plugins/mp.py
+++ b/nose2/plugins/mp.py
@@ -21,12 +21,31 @@ class MultiProcess(events.Plugin):
     def __init__(self):
         self.addArgument(self.setProcs, 'N', 'processes', '# o procs')
         self.testRunTimeout = self.config.as_float('test-run-timeout', 60.0)
-        self.procs = self.config.as_int(
-            'processes', multiprocessing.cpu_count())
+        self._procs = self.config.as_int(
+            'processes', 0)
         self.setAddress(self.config.as_str('bind_address', None))
 
         self.cases = {}
 
+    @property
+    def procs(self):
+        """Get the appropriate number of procs for self.procs if self._procs is
+        0."""
+
+        if self._procs == 0:
+            try:
+                self._procs = multiprocessing.cpu_count()
+            except NotImplementedError as e:
+                self._procs = 1
+        return self._procs
+
+    @procs.setter
+    def procs(self, value):
+        """Setter for procs property"""
+        if value <= 0:
+            raise AttributeError("Can't set the procs number to less than 1")
+        self._procs = value
+
     def setProcs(self, num):
         self.procs = int(num[0])  # FIXME merge n fix
         self.register()
```

Cheers,

P-EB:
Closes #381

### DIFF
--- a/nose2/plugins/mp.py
+++ b/nose2/plugins/mp.py
@@ -21,11 +21,30 @@ class MultiProcess(events.Plugin):
     def __init__(self):
         self.addArgument(self.setProcs, 'N', 'processes', '# o procs')
         self.testRunTimeout = self.config.as_float('test-run-timeout', 60.0)
-        self.procs = self.config.as_int(
-            'processes', multiprocessing.cpu_count())
+        self._procs = self.config.as_int(
+            'processes', 0)
         self.setAddress(self.config.as_str('bind_address', None))
 
         self.cases = {}
+
+    @property
+    def procs(self):
+        """Get the appropriate number of procs for self.procs if self._procs is
+        0."""
+
+        if self._procs == 0:
+            try:
+                self._procs = multiprocessing.cpu_count()
+            except NotImplementedError as e:
+                self._procs = 1
+        return self._procs
+
+    @procs.setter
+    def procs(self, value):
+        """Setter for procs property"""
+        if value <= 0:
+            raise AttributeError("Can't set the procs number to less than 1")
+        self._procs = value
 
     def setProcs(self, num):
         self.procs = int(num[0])  # FIXME merge n fix
--- a/nose2/plugins/mp.py
+++ b/nose2/plugins/mp.py
@@ -42,8 +42,8 @@ class MultiProcess(events.Plugin):
     @procs.setter
     def procs(self, value):
         """Setter for procs property"""
-        if value <= 0:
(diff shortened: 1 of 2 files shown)


ptthiem:
I wanted to hand test something, just to see if I'm being paranoid.  Once, I do that, I think this will be fine.

P-EB:
Thanks

nose2/plugins/mp.py
```diff
@@ -21,12 +21,31 @@ class MultiProcess(events.Plugin):
...
+            try:
+                self._procs = multiprocessing.cpu_count()
+            except NotImplementedError as e:
+                self._procs = 1
+        return self._procs
+
+    @procs.setter
+    def procs(self, value):
+        """Setter for procs property"""
+        if value <= 0:
```
ptthiem: If you do `nose2 --plugin nose2.plugins.mp -N`  this will throw an `AttributeError`.   I would say this should be `< 0`, so that the default works.  And probably adjust the error message.    I probably should write a test case for this scenario, we can just file a bug for that.  Otherwise, I think the patch is fine.

P-EB:
@ptthiem Indeed, my bad for the error. :)

katrinabrock:
commit merged in #389

PR #392, opened 2018-04-26
junitxml: Use test module in place of classname if no classname exists

rhelmot:
I noticed an issue that the xml output for ordinary function testcases has absolutely no indication for what file the test was written in, while the generator testcases have a classname marked as the name of the module. imo the latter is desirable behavior, so I've tried to replicate it for ordinary testcases.

For one of my testcases named `test_crosspage_read` in a file named `test_memory.py`, this PR changes the XML output from this:

```
 <testcase classname="" name="test_crosspage_read" time="0.026392">
   <system-out/>
 </testcase>
```

to this:

```
  <testcase classname="test_memory" name="test_crosspage_read" time="0.042382">
    <system-out />
  </testcase>
```

### DIFF
--- a/nose2/plugins/junitxml.py
+++ b/nose2/plugins/junitxml.py
@@ -130,6 +130,8 @@ class JUnitXmlReporter(events.Plugin):
 
         testcase = ET.SubElement(self.tree, 'testcase')
         testcase.set('time', "%.6f" % self._time())
+        if not classname:
+            classname = test.__module__
         testcase.set('classname', classname)
         testcase.set('name', method)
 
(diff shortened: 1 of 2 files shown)


katrinabrock:
Seems reasonable. Can you write a test for this?

rhelmot:
The answer to "but who will test the testers?" is apparently just "the testers"

Testcase pushed.

sirosen:
> but who will test the testers?

Indeed! Beyond the trope, there's a legitimate problem.
JUnit is meant for downstream consumers, but we don't know exactly what that consumer will be.

That is somewhat academic though: if someone is *expecting* `classname=""`, they're working around a bug in nose2. We shouldn't *avoid* fixing bugs out of fear.

Looks good to me. Merging.

PR #393, opened 2018-05-04
Add code to enable plugins to documentation

m-kuhn:
E.g.

```
[unittest]
plugins = nose2.plugins.logcapture
```

### DIFF
--- a/nose2/sphinxext.py
+++ b/nose2/sphinxext.py
@@ -71,7 +71,7 @@ class AutoPlugin(Directive):
 
         # config options
         if config.vars:
-            self.add_config(rst, config)
+            self.add_config(rst, config, plugin)
 
         # command-line options
         if opts.opts:
@@ -87,7 +87,7 @@ class AutoPlugin(Directive):
         rst.append(u'   :members:', AD)
         rst.append(u'', AD)
 
-    def add_config(self, rst, config):
+    def add_config(self, rst, config, plugin):
         headline = u'Configuration [%s]' % config.section
         self.headline(rst, headline)
 
@@ -104,6 +104,9 @@ class AutoPlugin(Directive):
                    u'the following in a :file:`unittest.cfg` file.', AD)
         rst.append(u'', AD)
         rst.append(u'.. code-block:: ini', AD)
+        rst.append(u'  ', AD)
+        rst.append(u'  [unittest]', AD)
+        rst.append(u'  plugins = %s' % plugin.__module__, AD)
         rst.append(u'  ', AD)
         rst.append(u'  [%s]' % config.section, AD)
         for var in sorted(config.vars.keys()):
--- a/nose2/sphinxext.py
+++ b/nose2/sphinxext.py
@@ -4,7 +4,7 @@ from docutils import nodes
 from docutils.statemachine import ViewList
 from docutils.parsers.rst import Directive, directives
 
-from nose2 import events, session, util
+from nose2 import events, session, util, plugins
 
 
 AD = u'<autodoc>'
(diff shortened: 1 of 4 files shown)


sirosen:
This is already covered by the [main plugins documentation page](http://nose2.readthedocs.io/en/latest/plugins.html). Some plugins are loaded by default, and some are not.

Plugins should which are loaded by default should not note the fact that you can write them down in unittest config -- it makes their docs longer without adding value.
However, every plugin which is not loaded by default should definitely have this line written in their docs. If it's missing for logcapture, we should add it.


We can either:
- close this and manually fix the docs for all of the not-loaded-by-default plugins

OR

- leave this, but refactor the [default plugin list](https://github.com/nose-devs/nose2/blob/2d550aa3f9770bfba7a5e2bd905a3f5d4850bdbc/nose2/main.py#L74-L88) into something which the sphinx extension can scan.

I'm inclined to just document it all by hand because it's simpler and less code to maintain. However, with the sphinx work already done, maybe it's worth considering making it part of the autodoc.
@katrinabrock, do you have a particular preference or see another option?

sirosen:
I think I mean the opposite of what you're saying.

By way of example, there's no point in having `logcapture` document that you can write
```
[unittest]
plugins = nose2.plugins.logcapture
```
That config does nothing, so adding it to the docs makes them longer, is potentially misleading (could make you think it's required), and adds no value.

It is, by contrast, relevant to document
```
[log-capture]
always-on = False
```
but that's already in the docs. Maybe it's insufficiently clear -- that's a simpler problem to solve.

It *does* add value to add this to plugins which are built-in but not loaded by default, for example:
```
[unittest]
plugins = nose2.plugins.junitxml
```

So we're you're going to inject mention of this conf into documentation for plugins automagically, I want to at least filter out `logcapture` and everything else that's loaded by default.

katrinabrock:
Ok, makes sense now. I haven't read all the gitter conversation yet, but based on the comments here this is what I'm thinking:

>That config does nothing
Ah, then for sure, let's not include it in doc.

>Maybe it's insufficiently clear -- that's a simpler problem to solve.

I think this is the real problem this PR is designed to solve.

Seems like the issue is "configuration options for default plugins not sufficiently/clearly documented", the two solutions in @sirosen first comment seem to address "configuration of non-default plugins not sufficiently/clearly documented". I'm just inferring this based on log capture being used as the example

>close this and manually fix the docs for all of the not-loaded-by-default plugins

I vote hard no on this. I want to avoid any situation where there's a "Every time we do X, we have to remember to do Y." if at all possible. In this case that's "Every time we add a not-loaded by default plugin, we have to remember to add a doc about the config that matches this template."

>leave this, but refactor the default plugin list into something which the sphinx extension can scan.

I think this is the way to go. Have a boiler-plate doc for default plugins and a different boilerplate doc for non-default plugins.  (For default something like "This plugin is enabled by default, do disable it, do X." and for non default "This plugin is not loaded by default, to enable it, do Y." where Y is more or less what's in this PR.)

I think this should be done before merge as IMO, because as @sirosen has described, the PR as-is will result in docs that are a bit misleading for default plugins.

sirosen:
I've gone ahead and pushed to this PR branch. @katrinabrock, assigning to you to review since making changes somewhat disqualifies me.

> I vote hard no on this. I want to avoid any situation where there's a "Every time we do X, we have to remember to do Y." if at all possible.

*shrugs* I get where you're coming from, and somewhat agree, but am also more than willing to compromise these principles if they get in the way of better docs. This time, certainly, the changes to the sphinx extension were pretty simple and easy.

nose2/sphinxext.py
```diff
@@ -88,6 +88,31 @@ def document(self, rst, plugin):
...
+        if plugin_modname not in plugins.DEFAULT_PLUGINS:
+            self.headline(rst, u'Enable this Plugin')
+            rst.append(
+                u'This plugin is built-in, but not loaded by default.', AD)
+            rst.append(u'', AD)
+            rst.append(
+                u'Even if you specify ``always-on = True`` in the '
+                u'configuration, it will not run unless you also enable it. '
+                u'You can do so by putting the following in a '
+                u':file:`unittest.cfg` or :file:`nose2.cfg` file',
```
m-kuhn: Would it be possible to link here to more information about where this file is searched?

m-kuhn:
Looks better to me than my approach 👍 
I lack the big picture of the nose2 code, so cannot comment too much on how clean it is.

nose2/sphinxext.py
katrinabrock: I would have to dig through to find the relevant code as well.

I'm going to merge this PR now, but feel free to open additional PR or open discussion on enhancing this text.

katrinabrock:
Looks good, merging.

Thanks @m-kuhn and @sirosen for your work on this.

PR #394, opened 2018-05-06
Add "prettyassert" plugin that gives bare asserts better support

### LINKED ISSUE #8: support bare asserts better
Something like the assert introspection in nose or py.test would make a nice plugin

sirosen:
Add "prettyassert" plugin, ~enabled~ loaded, but not enabled, by default.
Make `assert` failures pretty, like `nosetests -d`.

Call `nose2 --pretty-assert` or turn it on in config, and it will print
- the relevant exception statement
- the exception message if given
- values of all variables used in the assert

The implementation of `nosetests -d` was the primary inspiration for this.
I ruled out the py.test approach as too laborious to be worth doing, as it involves rewriting bytecode on the fly. Plus, there are all sorts of caveats given with that implementation and scenarios where it's known to blow up with other plugins, so we'd want this implementation as a fallback for those cases anyway.

This is not an exact reimplementation of the `nosetests -d` approach. I've tried to make some modest improvements to the behavior, aided in part by the fact that we don't have to support ancient python versions.

Closes #8

EDIT: fix "enabled" to "loaded, but not enabled"


ptthiem:
This looks pretty cool

sirosen:
This morning I thought of a case that this doesn't handle well:
```
import unittest


class TestFoo(unittest.TestCase):
    x = 1

    def test_ohnoez(self):
        x = 2
        assert self.x == x
```

gives

```
>>> assert self.x == x

values:
    self = <test_demo.TestFoo testMethod=test_ohnoez>
    x = 2
```

That's not terrible, and I don't know that I want to try solving attribute resolution in the first version of this plugin, but I know I've written tests for webapps like this:
```
self.assertEqual(response.status, 200)
```
so attribute resolution would be somewhat necessary to rewrite that as
```
assert response.status == 200
```

sirosen:
It's been a morning.
Refactored to do attribute resolution and handle certain cases which failed before, e.g.
```
assert x == (y
             ).z
```

These now resolve correctly, by adding lines to the relevant block until it parses.

There might be other obscure failure cases I haven't thought of or tested for.

The attribute resolution is pretty crude and can be deceived if you're being clever, but generally it will do the right thing and handles the simple cases.
When it fails hard, it will hopefully just print the original assert statement (as a whole) and leave it to the user. This is still better than the built-in python `assert` behavior for test cases.

katrinabrock:
>Closes #8

Nice!

Looks cool. I'm surprised you didn't need to change any existing tests. 

> enabled by default
Reserving judgement on this atm. I'll take a deeper look later this week if that's ok.

sirosen:
> enabled by default

Sorry, that was wrong.

It's *loaded* by default, but off. (So you don't have to put it into your plugins list).

I left it disabled because I'm a bit fearful that there's some crazy expression like
`assert True; assert False`
which will blow it up.

nose2/plugins/prettyassert.py
```diff
@@ -0,0 +1,360 @@
...
+        # okay, get repr() for a good string representation
+        strvalue = repr(obj)
+        # append in the form we want to print
+        collect_tokens.append('    {} = {}'.format(name, strvalue))
+
+    return assert_startline, collect_tokens
+
+
+class TokenProcessor(object):
+    def __init__(self, f_locals, f_globals):
```
katrinabrock: oh wait, `f` is for frame?

nose2/plugins/prettyassert.py
```diff
@@ -0,0 +1,360 @@
...
+
+        When an identifiable/usable token is found, add it to
+        collect_tokens
+        (which we add to extraDetail later)
+
+        When an "assert" statement is found, reset the token collection
+        and return the start line (relative to the text being tokenized)
+        """
+        prior_tok = self.last_tok
+        self.last_tok = (toktype, tok)
```
katrinabrock: shouldn't this either be set at the end or be called `current` tok?

nose2/plugins/prettyassert.py
```diff
@@ -0,0 +1,360 @@
...
+        # OP is operators, including .
+        #
+        # special note: don't clear resolution for whitespace (e.g. newline)
+        if toktype not in (tokenize.NAME, tokenize.OP):
+            # only newline for now, maybe we'll find others
+            if toktype not in (tokenize.NEWLINE,):
+                self.doing_resolution = None
+            return
+
+        """
```
katrinabrock: can you make these multi line comments consistent with either `"""` or `# `?

nose2/plugins/prettyassert.py
```diff
@@ -0,0 +1,360 @@
...
+            self.doing_resolution = None
+            return start[0]
+
+        # handle tokens
+
+        # CASE 3: a name is being resolved,
+        #         there is a previous token,
+        #         and it's a "." operator
+        if self.doing_resolution and prior_tok and (
+                prior_tok[0] == tokenize.OP and prior_tok[1] == '.'):
```
katrinabrock: maybe use a namedtuple for prior_tok? so reader doesn't have to scroll back an remember which is at index 0 vs 1.

nose2/plugins/prettyassert.py
```diff
@@ -0,0 +1,360 @@
...
+        reset resolution for everything else, e.g. "[", "]", ":"
+        special note: reset resolution for "("
+
+        failing to filter out ")" can result in badness in cases like this:
+            >>> def foo():
+            >>>     return [1]
+            >>> foo.pop = 2
+            >>> ...
+            >>> def test_foo():
+            >>>    assert foo().pop() == 2
```
katrinabrock: write a test for this case?

nose2/plugins/prettyassert.py
```diff
@@ -0,0 +1,360 @@
...
+            # we've never seen this token before
+            else:
+                # try to resolve to a value
+                try:
+                    value = self.f_locals[tok]
+                except KeyError:
+                    try:
+                        value = self.f_globals[tok]
+                    except KeyError:
+                        # unresolveable name -- short circuit
```
katrinabrock: What situations could lead to this and what is the overall outcome of returning here?

sirosen:
To answer some questions while I have the time...

> Overall design comment: [elided for brevity]

Yeah, I'll switch to a two-pass model of "data collection" followed by "rendering". That will probably lend itself better to debugging too, if we see issues (e.g. we can slap a repr on the AssertionVariableCollection object, or whatever its called).

> I'm not super familiar with inspect or tokenize so I can't comment on whether they're being used optimally.

I'm not really either. I've used inspect before, and IMO it's usually misapplied to enact one antipattern or another. This seems like a rare, legitimate use-case.
I suspect there may be subtle issues with this usage of tokenize, which is why I'm trying to be ready for it to bomb out in a few places.

> Made a bunch of comments on areas where code could be more clear.

Yeah. I'll try to tackle all of these when I do the rework.

> Testcases look like a good start, but might be good to add a few more. For example, what happens if the unittest assert methods e.g. the one you cite self.assertEqual(x, y) are use with this plugin. Is the output altered in any way? Does the plugin work in the case `assert x == y == z?

Upon closer inspection, I think this needs more testing around the unittest `assert*()` methods.
It turns out those raise custom AssertionErrors, so they'll be captured and behave oddly.

Maybe dropping support for `raise AssertionError(...)` is the right thing to do -- we can try to look for the `assert` token and act as a no-op if we don't find it.
There will probably be odd subtle cases like
```
raise AssertionError('foo'); assert x == y
```
which would get tripped up, but that's okay in my book.

> If there are situations where you already know it breaks, (like assert True; assert False, based on your comment). Would be good to add a FIXME comment. Also, it would be nice for the plugin realize something is wrong and fall back on non-pretty behavoir. (If this is in fact what happens, can you add a test of it?)

I wasn't able to come up with any scenarios where things *break*, but if you think of odd ones, let me know. They're all worth simply dumping in the testsuite (since these particular tests are very fast).
The `assert x; assert y` case, mentioned earlier, is definitely one I should add.

> Also, have you tried running all the nose2 tests with this option enabled? Just curious if anything breaks.

Yeah, nothing breaks, but it's not clear how much that means. I doubt our tests are using bare asserts anywhere, *plus* all this stuff fires on failures, so the plugin code is basically a no-op all the time.

> Not sure how much more time you want to spend on this. I think at a minimum you should add a few more tests before merging. How much you want to refactor for readability is up to you.

It's definitely worth it, but I haven't been able to carve out time. I'll update again (and reply to various comments as appropriate) when I can.

nose2/plugins/prettyassert.py
sirosen: The outcome is that we skip this value in the output.
Simple example scenario:
```
x = 1
assert x == 0; y = 1
```

We'll hit `y` and not have a value for it in the frame. It should be omitted from output.

I'll add a test case for that when I get to all of this (proving harder than expected).

sirosen:
I know this has been lingering for quite a while, as I kept finding it difficult to carve out time to do that second-pass over it that was wanted, but I've just done so and think it's in "okay" shape, at least.

There are a couple of questions I still have, in my own mind, about this:
- Should `AssertionError`s which don't originate from an `assert` statement -- or at least, not one that we can find in trying to tokenize things, e.g. `exec('assert False')` -- still get caught and treated? That would make this capture unittest assertion methods, which is kind of weird (but presumably you use one or the other?)
- Does this need a bunch more tests? (Probably yes, but it's late in the day here, so closing up for now and will try to do more on this.)

sirosen:
I've changed the behavior on unittest assertion methods, so that they are not altered in any way. I think this is the correct/expected behavior -- only AssertionErrors originating in an `assert` statement should be decorated.

Along with the last couple of tests that I've added, I can't think of anything else in this that really ought to change. As above, more testing is always possible, but I can't think of anything which really needs to be added.

sirosen:
Although this benefited from and was improved by review, I want to release it as part of v0.9.0 and am therefore merging it.

PR #406, opened 2018-08-14
support test labels in nose2, can choose tests at runtime by specifiying matching labels using --labels arg

tupakulasuresh:


sirosen:
I think that this work duplicates the existing [attributes plugin](https://nose2.readthedocs.io/en/latest/plugins/attrib.html).

Is there a significant distinction between the two? If there is, can this possibly be refactored into an option for the attributes plugin? They seem very similar.

tupakulasuresh:
Thanks @sirosen. I didn't check the attributes plugin. I will explore little more. 
closing this PR

PR #410, opened 2018-08-15
Support verbosity set in config, count `-q`

### LINKED ISSUE #310: verbosity is ignored when set in configuration file or as an argument to nose2.main()
This is, I think, related to, but is not the same as, issue #115. After seeing no affect when I set verbosity in my .cfg file or passed it as an argument from nose2.main(), I used a debugger to dive into the code and it seemed to lose its verbosity somewhere along the line and revert to using the session's default verbosity of 1.

I haven't been able to figure out what value verbosity takes when nose2 is run from the command line with "-v", though that does work.

sirosen:
- add `--verbosity` for explicitly setting the base verbosity level as an int
- add counting of `-q` option and subtract `-q`s from `-v`s
  - You can get the old behavior with `--verbosity 0`
- allow `verbosity = <int>` in config
- cleanup verbosity handling logic -- it all belongs to the session now
- change start_dir logic to follow model of verbosity logic

This is a work-in-progress, but I'm opening the PR right away with the initial work so that it can get earlier review.
It still needs additional tests for the new/altered behaviors, and additions to the 0.8.0 changelog.

This will close #310 and #115.

### DIFF
--- a/nose2/main.py
+++ b/nose2/main.py
@@ -155,9 +155,17 @@ class PluggableTestProgram(unittest.TestProgram):
             dest='exclude_plugins', default=[],
             help="Do not load this plugin module")
         self.argparse.add_argument(
-            '--verbose', '-v', action='count', default=0, help="print test case names and statuses")
-        self.argparse.add_argument('--quiet', action='store_const',
-                                   dest='verbose', const=0)
+            '--verbosity', type=int,
+            help=("Set starting verbosity level (int). "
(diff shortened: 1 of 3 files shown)


sirosen:
Removing the "WIP" from this. I just added some tests and it seems to work correctly. Assuming it passes on Travis, I plan to self-merge for inclusion in 0.9.0.

PR #412, opened 2018-09-09
junit-xml: log skip reason

marmarek:


sirosen:
Sorry for the slow reply. I'm going to go ahead and merge, since this is quite harmless, and open up a backlog issue to add a test for this behavior.

I'd love it if you could add said test, but I don't think this is worth blocking over that.

PR #414, opened 2018-09-17
Test skip reason in junixml plugin

### LINKED ISSUE #413: Add test for junit-xml output containing skip reason
#412 made junit-xml output include the skip reason, and omits those attributes when no reason is given.

It would be nice, to follow the rule of testing for every change, to add a test just for that behavior.

marmarek:
Fix #413


sirosen:
Thanks so much for doing this! 👍 

Took a while for me to come back to `nose2` and try starting to clean up the open issues, hence the delay. The test looks good.

PR #415, opened 2018-09-24
setup.py: Add Python 3.7 trove classifier

cclauss:
Upgrade the Travis CI version of Python 3.7 from prerelease to production Python 3.7 in alignment with travis-ci/travis-ci#9069 and then add the Python 3.7 trove classifier in setup.py.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,12 @@ python:
 - '3.4'
 - '3.5'
 - '3.6'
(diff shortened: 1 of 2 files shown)


sirosen:
Thanks for adding this. I'd also like to add 3.8-dev and python nightly testing with failure allowed, but I'll do that in a separate follow-up.

PR #416, opened 2018-10-09
Update the requirements: coverage and six 

cclauss:
Also, simplify by using [__six.unichr()__](https://six.readthedocs.io/#six.unichr).

### DIFF
--- a/nose2/plugins/junitxml.py
+++ b/nose2/plugins/junitxml.py
@@ -277,7 +277,7 @@ ILLEGAL_REGEX_STR = \
     six.u(']')
 RESTRICTED_REGEX_STR = \
     six.u('[') + \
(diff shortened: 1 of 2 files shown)


sirosen:
Although switching to `six.unichr()` seems like a positive change, I'm hesitant to tighten the requirements for `nose2` without careful consideration.
The reason is that `nose2` will typically need to be installed into the same environment as the package which it is testing.

If someone has a package with `six==1.10.0` as a requirement, they can currently use `nose2` to test it without issue. If the constraint tightens to `six>=1.11.0,<2` or similar, testing for that package will break.

For that reason, I plan to rebase and accept the change to use `six.unichr()`, a function which was added in `six` v1.4, but reject the other changes here.
If there's a strong reason for `nose2` to require newer versions of these libs, just say the word and I'm happy to reconsider and/or discuss.

PR #420, opened 2018-11-05
Remove redundant code and add python_requires

hugovk:
* Remove some redundant code and comments for already-dropped EOL versions

* Add `python_requires` to help pip install the correct version for a user's Python


sirosen:
I'd rather leave out `python_requires` because I think it does more harm than good for us.

You can install `nose2` on py3.2, but we can't guarantee that it works because that's an EOLed python which we don't test. Things might work fine -- the project just doesn't try to make promises about that.

@hugovk, would you be alright with dropping 22ca2bd from this, and just merging with the other commit? I'm more than happy to take in those other changes -- die py2.6 comments, die! -- and thanks for doing this. I think Travis builds also need some adjustment... I'll do that myself soon.

hugovk:
Sure, https://github.com/nose-devs/nose2/commit/22ca2bdc57c6e6cc881296bc3524f9896813b291 is dropped!

sirosen:
Sorry, I forgot about this. Thanks again, merging now.

PR #425, opened 2018-12-14
Ensure only regular files are tested

mnm678:
The project currently loads any file with a .py extension. This ensures that the files loaded are regular files and not pipes, fifos, or anything else that will cause the application to hang.

### DIFF
--- a/nose2/plugins/loader/discovery.py
+++ b/nose2/plugins/loader/discovery.py
@@ -140,7 +140,7 @@ class Discoverer(object):
             for test in self._find_tests_in_dir(
                 event, full_path, top_level):
                 yield test
-        elif os.path.isfile(start):
+        elif os.path.isfile(start) and stat.S_ISREG((os.stat(start)).st_mode):
             for test in self._find_tests_in_file(
                 event, start, full_path, top_level):
                 yield test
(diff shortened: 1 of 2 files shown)


mnm678:
I added a test, but can't seem to add the support directory to git. Is there a way to generate that?

sirosen:
Sorry it's taken me so long to circle around on this to review.

I think this patch is fine.
My big worry at first was that this could cause new issues with symlinks, broken symlinks, but I'm unable to come up with any scenarios which are made worse by the addition of this check.

That said, why do you need this? What is your situation such that you're generating pipes named `.py`? Is it really necessary?

I'd like to understand what we're fixing here before accepting this, and to put in a comment describing it.
Otherwise it's equally possible that someone will show up tomorrow and want to revert the change, and we won't have a rationale for accepting this and refusing that.

> I added a test, but can't seem to add the support directory to git. Is there a way to generate that?

I see that the dir for your test is missing, but I'm not sure exactly what you're asking. There's no tooling for generating those directories.

If you have a dir in `nose2/tests/functional/support/scenario/`, you should be able to add it just fine. If you can't, you may have blacklisted it in your global gitignore settings, and could force it with a `git add -f`.

sirosen:
As it's a year since last touch, I'm going to close this.

I'm happy to discuss further and even accept the change, but I'm not willing to do so without an understanding of why it's needed.

Issue #426, opened 2019-01-31
multiprocess plugin hang for the event can not be serialized.


SexyC:
Hi nose2 experts,

I found a problem in the mp plugin that if the event sent cannot be serialized, the `conn.send((testid, events))` will throw exception and cause the process quit. Thus this will exhaust the worker processes and cause the whole tests hang.

```python
def procserver(session_export, conn):
    # init logging system
    rlog = multiprocessing.log_to_stderr()
    rlog.setLevel(session_export['logLevel'])

    # make a real session from the "session" we got
    ssn = import_session(rlog, session_export)

    if isinstance(conn, collections.Sequence):
        conn = connection.Client(conn[:2], authkey=conn[2])

    event = SubprocessEvent(ssn.testLoader,
                            ssn.testResult,
                            ssn.testRunner,
                            ssn.plugins,
                            conn)
    res = ssn.hooks.startSubprocess(event)
    if event.handled and not res:
        conn.send(None)
        conn.close()
        ssn.hooks.stopSubprocess(event)
        return
    # receive and run tests
    executor = event.executeTests
    for testid in gentests(conn):
        if testid is None:
            break
        # XXX to handle weird cases like layers, need to
        # deal with the case that testid is something other
        # than a simple string.
        test = event.loader.loadTestsFromName(testid)
        # XXX If there a need to protect the loop? try/except?
        rlog.debug("Execute test %s (%s)", testid, test)
        executor(test, event.result)
        events = [e for e in ssn.hooks.flush()]
        conn.send((testid, events))       # <== May throw exception and quit the process
        rlog.debug("Log for %s returned", testid)
    conn.send(None)
    conn.close()
    ssn.hooks.stopSubprocess(event)
```

I'd suggest we catch the exception and log it for debugability and keep the process continue to do other works.
For example.
```python
try:
    conn.send((testid, events))
except:
    rlog.exception("Fail sending event %s (%s)" % (testid, test))
```

Does that make sense to you?

Thanks!
Best Regards.

ptthiem:
Good catch.  That should be an easy fix.

PR #429, opened 2019-02-11
For issue #426. Fix the issue that multiprocess plugin hang for the event can not be …

SexyC:
To fix issue #426.

When the event sent to main process can not be serialized, conn.send
will throw out exceptions. The worker process will exit abnormally. And
the main process will also hang at the conn.recv. So the entire unit
tests run got hang.

This change fixes this issue by catching the exception that conn.send
may throw out, log it for debugability and send empty event list to
unblock the conn.recv.


SexyC:
Hi nose2 experts,

May I ask a stupid question?

When will this change be released to the pip repository? I saw the pip repository still used the nose2 v0.8.0.

Any way we can use the latest nose2 via pip install?

Thank you very much!
Best Regards.

PR #449, opened 2019-09-04
Enable multiprocess compatibility within coverage plugin

bugfood:
Hi,

I found that when I used both `--plugin=nose2.plugins.mp` and `--with-coverage` that the coverage report showed much lower coverage than when `mp` was not in use.

I wrote up a patch to fix things. Mostly, this consists of:
1. Make `coverage` plugin subprocess-aware and start/stop coverage within subprocesses.
2. Enable writing to coverage files named with a suffix; combine those files before reporting.

There are a couple caveats which are documented in `docs/plugins/coverage.rst`. The first caveat (command-line options ignored) is disappointing, but I didn't see a way to fix this; `handleArgs()` is not invoked from within `mp`. This is my first time using nose2, though, so maybe I'm missing something; if someone can give me a tip on how to fix that caveat, I can attempt to do so.

I also included a patch to tidy up the `mp` documentation a bit.

Thanks,
Corey

### DIFF
--- a/docs/plugins/coverage.rst
+++ b/docs/plugins/coverage.rst
@@ -18,3 +18,18 @@ Otherwise, please be aware of the following known differences:
 
 - The ``fail_under`` parameter results in an exit status of 2 for ``coverage``,
   but an exit status of 1 for ``nose2``
+
+Compatibility with mp plugin
+----------------------------
+
+The ``coverage`` and ``mp`` plugins may be used in conjuction to enable
+multiprocess testing with coverage reporting.
+
+Special instructions:
+
+- Due to the way the plugin is reloaded in subprocesses, command-line options
+  for the ``coverage`` plugin have no effect. If you need to change any
+  ``coverage`` plugin options, use a configuration file.
+- Do *not* use the ``concurrency`` option within a ``.coveragerc`` file ; this
+  interferes with the ``coverage`` plugin, which automatically handles
+  multiprocess coverage reporting.
--- a/docs/plugins/mp.rst
+++ b/docs/plugins/mp.rst
@@ -44,9 +44,9 @@ config file::
    defaults to the number of CPUs available. Also note that a value of 0 will
    set the actual number of processes to the number of CPUs on the computer.
 
-Should one wish to specify the use of internet sockets for 
+Should one wish to specify the use of internet sockets for
 interprocess communications, specify the ``bind_address``
-setting in the ``[multiprocess]`` section of the config file, 
+setting in the ``[multiprocess]`` section of the config file,
 for example::
 
   [multiprocess]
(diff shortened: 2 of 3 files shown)


sirosen:
Thanks for the pull request. I have to read this carefully and test it out a bit, but after a cursory review this looks good.

bugfood:
Thanks @sirosen . I'm keeping an eye on the PR and if you have any feedback I'll address it when I can.

ptthiem:
I did some poking around to refresh my memory.  I doubled checked a note about setting related, but I am pretty sure that if we set the self.config variables in HandleArg (just a reference to a safeparser section), it will get passed to the child process.  The child shouldn't be rereading the file; the whole config object is exported from the session.

ptthiem:
Ok.  I mispoke, it looks like we're copying it. Hmm.  I wonder if we really need to.  Perhaps. Self.session.config. That is the global config safeconfigparser.

bugfood:
Thanks @ptthiem , I see you have a lot of commits to mp.py.

I remember looking at the `handleArgs` situation and knowing that the subprocesses are copies of the parent and have that data in memory somewhere, but just without access to it. I didn't really have any immediate ideas for how to pass that data, but one thing that stuck in my head is that I didn't know of a reason for why the session parameters are serialized and sent to each subprocess via the pipe/connection. I don't even know if the subprocesses really need to make a new session; maybe they can mutate the old one to fit their needs.

In any case, I took a stab at what seems to me to be the first step: passing the `MultiProcess` object itself to `procserver()`. At that point, the session parameters can be copied directly, and this approach might open doors that facilitate easier data passing for other purposes:
https://github.com/nose-devs/nose2/pull/450

ptthiem:
They must be serialized on windows.  Windows doesn't not fork.

bugfood:
Dang, I was afraid of that.

bugfood:
@sirosen @ptthiem is there anything blocking merge of this patch?

My impression is that, while it doesn't make everything perfect, it's still an improvement without any regression.

I rebased the PR to current master.

Thanks,
Corey

sirosen:
Hi, I'm sorry to have dropped the ball on this. I'll try to look at it later tonight (US Eastern TZ) to try to determine whether or not there's any reason not to merge it. I became hesitant given that Philip's feedback pertained to things that I'm not that knowledgeable about.

Thanks for rebasing, as it keeps things clean.

ptthiem:
Regarding my comments.  This is probably fine as is.  It was his response to me which lead to a second PR (which he closed) that started a deeper discussion. Sorry for the confusion.  My first set of comments was the mistaken recollection that if one were to just use HandleArgs to update the plugin's config object then that would fix his first caveat, instead he'd have to directly edit the global session config.  I can debate how to address that later.

sirosen:
Thanks for chiming in on this to clarify. Rereading the thread, I'm pretty sure that I misunderstood in the past.

I'm going to merge this right now, and I have a test I was just writing which passes on my machine, but which I'm going to put into a PR to ensure it works on Travis.

bugfood:
Thank you very much!

PR #450, opened 2019-09-19
mp: initiate new session based directly on old session

bugfood:
This takes advantage of the fact that a subprocess is a full copy of the
parent process, thus it has access to all the original data.

Rather than serializing the session information and passing it through
the pipe/connection, simply pass the original MultiProcess object to
procserver() as an argument.

Cons:
* Somewhat blurs the interface between a subprocess and its parent.

Pros:
* Less code.
* Eliminates the caveat that classes must be pickeable.

Continuing on this path should allow for parent-->child data passing
that is not currently available.

### DIFF
--- a/nose2/plugins/mp.py
+++ b/nose2/plugins/mp.py
@@ -26,6 +26,7 @@ class MultiProcess(events.Plugin):
         self.setAddress(self.config.as_str('bind_address', None))
 
         self.cases = {}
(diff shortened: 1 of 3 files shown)


bugfood:
According to:
https://github.com/nose-devs/nose2/pull/449#issuecomment-533094047
...this isn't portable to windows. Oh well, it was worth a try.

PR #456, opened 2019-12-05
Use collections.abc for Python 3.9 compatibility.

tirkarthi:
* Importing ABC from collections module was deprecated and removed in Python 3.9 . Use `collections.abc` instead for Python 3.9 compatibility. Reference : https://github.com/python/cpython/pull/10596

### DIFF
--- a/nose2/plugins/mp.py
+++ b/nose2/plugins/mp.py
@@ -3,7 +3,10 @@ import logging
 import multiprocessing
 import select
 import unittest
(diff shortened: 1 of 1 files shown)


sirosen:
Yeah, this seems reasonable.
I'll see about adding 3.9 as an "allow failures" part of the build matrix at some point. Our Travis setup needs a bit of a refresh.

Issue #465, opened 2020-03-27
mp plugin has issues with unittest.skip


twizmwazin:
I seem to have come across a weird issue when using `unittest.skip` on individual methods in a `unittest.TestCase` in combination with the `nose2.plugins.mp` plugin. Here is an example to show my issue:

Code (`test_ex.py`):
```py
import unittest

class MyTest(unittest.TestCase):
    @unittest.skip("aaa")
    def test_something(self):
        pass
```

Running without `mp` (correct behaviour):
```
$ nose2 -s . -v
test_something (test_ex.MyTest) ... skipped aaa

----------------------------------------------------------------------
Ran 1 test in 0.000s

OK (skipped=1)
```

Running with `mp`:
```
$ nose2 -s . -v --plugin nose2.plugins.mp -N1
test_not_found (nose2.loader.LoadTestsFailure) ... skipped aaa

----------------------------------------------------------------------
Ran 1 test in 0.005s

OK (skipped=1)
```

And one more example, if I comment out the `unittest.skip` decorator:
```
$ nose2 -s . -v --plugin nose2.plugins.mp -N1
test_something (test_ex.MyTest) ... ok

----------------------------------------------------------------------
Ran 1 test in 0.005s

OK
```

When the mp plugin is enabled and the skip decorator is used, the test is incorrectly labelled. In addition to the misreporting, this also is an issue when naming specific tests to run:

```
$ nose2 -s . -v --plugin nose2.plugins.mp -N1 test_ex.MyTest.test_something
test_not_found (nose2.loader.LoadTestsFailure) ... ERROR

======================================================================
ERROR: test_not_found (nose2.loader.LoadTestsFailure)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kevin/.virtualenvs/angr/lib/python3.7/site-packages/nose2/util.py", line 165, in try_import_module_from_name
    module = __import__('.'.join(splitted_name))
ModuleNotFoundError: No module named 'test_ex.test_something'; 'test_ex' is not a package

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/kevin/.virtualenvs/angr/lib/python3.7/site-packages/six.py", line 693, in reraise
    raise value
  File "/home/kevin/.virtualenvs/angr/lib/python3.7/site-packages/nose2/plugins/loader/functions.py", line 81, in loadTestsFromName
    result = util.test_from_name(name, module)
[...]

PR #466, opened 2020-04-08
test_name: Fix name conversion for UnitTest methods.

ltfish:
For UnitTest-wrapped functions, test_name() would convert a
module.class.test_function into "module.test_function" and lose the
class information. This is, unfortunately, not expected by the
MultiProcess plugin, which has a different way of converting tests to
their testids. This fix ensures for UnitTest-wrapped functions,
test_name() returns a "module.class.function"-style name that complies
with the manner that MultiProcess plugin expects.

This PR fixes issue #465.

### DIFF
--- a/nose2/util.py
+++ b/nose2/util.py
@@ -183,9 +183,12 @@ def test_name(test, qualname=True):
     if hasattr(test, '_funcName'):
         tid = test._funcName
     elif hasattr(test, '_testFunc'):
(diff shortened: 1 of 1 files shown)


sirosen:
With #472 merged, I believe this should now be closed as no longer necessary/desirable, since the bug has been fixed at a "deeper" level.

@ltfish, although I'm closing this, I want to say a big thanks for submitting this fix!

PR #469, opened 2020-06-08
Ignore top-level setup.py when scanning for doctests.

### LINKED ISSUE #460: Enabling [doctest] always-on = true makes nose2 exit with "error: no commands supplied"
I'm trying to migrate check-manifest's test suite from nose to nose2.  I have a couple of doctests I'm trying to include in the test suite, by creating a `unittest2.cfg` with

```ini
[unittest]
plugins = nose2.plugins.doctests

[doctest]
always-on = true
```

However as soon as I do that, running `nose2` with no command-line arguments fails:

```
$ nose2
usage: nose2 [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: nose2 --help [cmd1 cmd2 ...]
   or: nose2 --help-commands
   or: nose2 cmd --help

error: no commands supplied
```

I can comment out the `always-on = true` line and then nose2 runs all the tests (but no doctests).

stefanholek:
Fixes issue where the doctest loader would try to import a package's setup.py and blow up.

This fixes #460 for the common case.

### DIFF
--- a/nose2/plugins/doctests.py
+++ b/nose2/plugins/doctests.py
@@ -42,6 +42,9 @@ class DocTestLoader(Plugin):
             return
 
         name, package_path = util.name_from_path(path)
+        # ignore top-level setup.py which cannot be imported
+        if name == 'setup':
+            return
         util.ensure_importable(package_path)
         try:
             module = util.module_from_name(name)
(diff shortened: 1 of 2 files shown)


sirosen:
At first, I was concerned that this would fail on cases like `foo/setup.py` (where `setup` is actually a valid module name).
However, looking at `name_from_path`, I don't think that's a concern. In that case, `name` would be `foo.setup`, is that right? I just want to make sure I understand the fix before merging.

stefanholek:
Yes, exactly.

PR #470, opened 2020-06-10
mp: Get the correct class name for test classes.

ltfish:
Test classes are wrapped around by the MethodTestCase class in
plugins/loader/testclasses.py. Unwrap them before retrieving the module
and the class name of the actual test class in the mp plugin.

This PR fixes the issue that the mp plugin fails to load any test cases that are inside a UnitTest class. My current check on `test.__class__.__name__` is hacky - I am open to better ideas!


sirosen:
As I just reviewed and merged #483, I'm going to close this now, but it was the basis for that work.

@ltfish, I'll give you attribution in the changelog for the next version, so that your contribution is noted in some way.

Issue #444, opened 2019-06-13
Multiprocess plugins incompatible with mocks or docktests?


FurcyPin:
Hi,

I have a python projects that uses doctest for some tests and mocks for some other tests, and I use nose2 to run everything.

It works like a charm, but when I tried to enable the `nose2.plugins.mp` plugin for MultiProcess
testing I got error messages on all my tests that were using either doctest or mocks.

For test using mocks, I got errors that look like this:
```
======================================================================
ERROR: test_not_found (nose2.loader.LoadTestsFailure)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.6/unittest/mock.py", line 1179, in patched
    return func(*args, **keywargs)
TypeError: test_method() missing 1 required positional argument: 'mock_settings'
```
The test looks like this
```
    @mock.patch('dynaconf.settings.get', side_effect=test_settings)
    def test_method(self, mock_settings):
        "test something"
```

For test using doctest, I got errors that look like this:
```
======================================================================
ERROR: _my_function (some.Object)
Doctest: some.Object._my_function
----------------------------------------------------------------------
TypeError: _my_function() missing 3 required positional arguments: 'a', 'b', and 'c'
```
Where my doctest looks like this:
```
class Object():

    def _my_function(self, a, b, c):
        """

        Doctest:
        >>> Object()._my_function('a', 'b', 'c')
        hello

        :param a:
        :param b:
        :param c:
        :return:
        """
        print("hello")
```

If I disable the `nose2.plugins.mp`, everything works fine though.

stefanholek:
The mock.patch issue has been fixed in PR #472.

As for doctests: No, they don't work with the multiprocessing plugin.

PR #472, opened 2020-07-27
Ignore test methods even when of types.FunctionType.

### LINKED ISSUE #468: Failed to run a specific test with mock.patch decorator using nose2
This is my test file `test_mock.py` that minimally reproduces the failure:

```
import unittest
from unittest.mock import patch

class TestMock(unittest.TestCase):

    @patch("urllib3.PoolManager")
    def test_foo(self, mock_urllib3_poolmanager):
        pass
```

When I use nose2 to run all tests in this file. Everything works.

```
bash-4.2# nose2 test_mock
.
Ran 1 test in 0.102s

OK
```

But when I use nose2 to select a specific test to run. It throws an error.

```
bash-4.2# nose2 test_mock.TestMock.test_foo
E
ERROR: test_mock.transplant_class.<locals>.C (test_foo)
  File "/var/lang/lib/python3.7/unittest/mock.py", line 1256, in patched
    return func(*args, **keywargs)
TypeError: test_foo() missing 1 required positional argument: 'mock_urllib3_poolmanager'

Ran 1 test in 0.087s

FAILED (errors=1)
```

Any insight is greatly appreciated. Thanks!

stefanholek:
Stop functions loader from accepting class methods even when they are decorated (with e.g. mock.patch) and thus of types.FunctionType.

This fixes #468 and half of #444.

### DIFF
--- a/nose2/plugins/loader/functions.py
+++ b/nose2/plugins/loader/functions.py
@@ -87,6 +87,7 @@ class Functions(Plugin):
 
         parent, obj, name, index = result
         if (isinstance(obj, types.FunctionType) and not
+            isinstance(parent, type) and not
             util.isgenerator(obj) and not
             hasattr(obj, 'paramList') and
             util.num_expected_args(obj) == 0):
--- a/nose2/tests/unit/test_functions_loader.py
+++ b/nose2/tests/unit/test_functions_loader.py
@@ -1,4 +1,11 @@
 import unittest
+
+try:
+    from unittest import mock
+except ImportError:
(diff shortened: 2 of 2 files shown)


stefanholek:
Appears to fix #465 as well. Nice!

sirosen:
This is one of several good changes that I should have merged in long ago.
I'm collecting them up and we'll do a release.
Apologies to everyone for the long wait, but I hope we can feel good knowing a release is around the corner. 😅

ltfish:
Thanks for merging! I think we can eventually switch off of our own fork.

PR #476, opened 2020-07-28
Update test functions id, str, and repr.

### LINKED ISSUE #475: Test functions string repr contains 'transplant_class.<locals>.C'
The internal `transplant_class` becomes visible when running test functions with nose2 -v.

stefanholek:
Make id and repr for test functions work the same as for test classes, generators, and parameterized tests.
    
Fixes issue where `transplant_class.<locals>.C` was printed as part of a FunctionTestCase's string representation. Also makes test.id() work.

Fixes #475.

### DIFF
--- a/nose2/plugins/loader/functions.py
+++ b/nose2/plugins/loader/functions.py
@@ -140,6 +140,14 @@ class Functions(Plugin):
             return tests
         else:
             case = util.transplant_class(
-                unittest.FunctionTestCase, obj.__module__)(obj, **args)
+                FunctionTestCase, obj.__module__)(obj, **args)
             tests.append(case)
         return tests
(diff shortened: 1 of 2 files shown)


stefanholek:
The failing check is Travis' fault: "Failed building wheel for cryptography" 🤷‍♂️ 
Will probably fix itself when triggered again.

sirosen:
I triggered a re-run of that job and it seems to be stuck again, and something's messed up with the `master` build as well. I'm testing it with a manual tox run right now while I read through things, and I'll look into the build issues this weekend.

sirosen:
Okay, my local run of tox is all passing, and I like the implementation. 👍

PR #485, opened 2021-01-24
session plugin registry

stefanholek:
Fixes #483


sirosen:
I'm going to merge this, but note that it might autoclose 483 (typo in the comment above). If so, I'll just reopen that.

PR #489, opened 2021-01-24
Some fixes for slow nose2 tests

### LINKED ISSUE #488: Small fixes for tests
- test_conn_accept waits far too long
- 127.1.2.3 is not a valid address on macOS 🤷‍♂️

stefanholek:
Fixes #488


sirosen:
Yeah, I was setting up GitHub Actions this weekend (PR coming today) and I saw this test was failing. I just set it up to skip, because I didn't want to dig into it yet, but if this solves it that would be awesome. I'll be rebasing my CI work to see if this sorts it all out.

PR #490, opened 2021-01-25
Switch to GitHub Actions for CI builds; get builds passing

sirosen:
This removes Travis config and introduces a GitHub Actions workflow.
As it does so, it specifies a matrix which includes windows and macos builds, and which allows windows builds to fail because tests are broken on Windows right now.

The Travis builds were broken and I wanted to make this switch over anyway.

Windows-specific workarounds are removed: there is now only one tox config and the `unit_workaround.py` file is gone. It seems the main purpose of that file was to invoke `TestProgram(module=None)`, but this is already done by `nose2.discover`. I don't believe that the file was aiding us in Windows compatibility; just making things more confusing.

There's minor cleanup to the coverage config, and I've renamed the `doc` extra to `dev` and added `coverage` to it. So it's a single "nose2 development" extra which gets any testing and doc building dependencies. Since that extra is internal, I'm not worried about renaming it.


sirosen:
The build was marked as failing as a result of the windows jobs which are failing. It turns out `continue-on-error` [doesn't stop the job from being marked as a failure](https://github.com/actions/toolkit/issues/399), so I've switched to ignoring the failure of the "run tests" step. That seems to work, but it also may be confusing because it makes it look like the windows tests are passing.

I'll work on getting windows tests to pass at some point soon -- even if it means adding `skipIf(on_github_actions() and os.name == 'nt', "fails on windows")` to tests, so that the workaround can be dropped.

.coveragerc
```diff
@@ -1,8 +1,13 @@
 [run]
 branch = True
+omit = nose2/tests/*
 source = nose2
 include = nose2/*
```
stefanholek, line 5: I think `source` and `include` are mutually exclusive. At least I got a warning at some point. Maybe drop `source` like you did in the report section?

stefanholek (changes requested):
Please add py38 and py39 to the test matrix. 

`test_run_with_mp` fails under py38+ on macOS which _may_ be due to ["on macOS, the spawn start method is now used by default in multiprocessing"](https://www.python.org/downloads/release/python-387/).

I don't know anything about coverage + mp so I have to pass on this one.

sirosen:
> Please add py38 and py39 to the test matrix.

I was going to save that for follow-up, but sure, I can do it now. 😄 

In the absolute worst case for the macos issue, we can add `@skipIf(platform.system() == "Darwin" and (3,8) <= sys.version_info < (3,9))`, so I'm not too worried about it gumming up the works. I'd rather have a full matrix with some things skipped or ignored than not have the tests running at all.

.coveragerc
sirosen: Good catch. I'm pretty sure we want `include`, not `source`. I'll double check the coverage docs and update.

stefanholek (approved):
Thanks

PR #492, opened 2021-01-31
Apply precommit for black, flake8, and isort linting

sirosen:
I actually did this early last year and pushed it to my fork, but never to the main repo.
I didn't want to make it harder to contribute, so I waited for the dust to settle on recent contributions and rebased, did cleanup, and applied the necessary tweaks.

After this, we have
- working pre-commit hooks with [`pre-commit`](http://pre-commit.com/); configurable with `pre-commit install`
- `tox -e lint` runs `pre-commit`; easily invoked as `make lint`
- all code passes flake8 linting
- black and isort for style enforcement
- no more invalid escape sequences in strings; use raw strings for regex patterns and similar usages
- github workflow config runs `tox -e lint` in one of the linux builds

The diff is huge, but it's 90% a combination of whitespace changes and `'` -> `"`.
Tests pass and there are no functional changes.
I'm not expecting line-by-line review (though you are free to do so if the 4000 line diff isn't intimidating); mostly a sanity check and signoff on introducing pre-commit and doing so in this way.


stefanholek (commented):
This is the first time I use the pre-commit thingy. I was afraid it would reject commits unless they conform, but instead it appears to apply black/flake/isort fixes before committing. Ok.

So first question: Should pre-commit be a [dev] dependency or is it supposed to be installed as a global utility like tox and twine? Should it go into the Makefile then?

As a new contributor how do I setup my local environment? Via the Makefile? If I never do `pre-commit install` no hooks will ever run, right?

We should document this and/or have something like `make develop`. Or am I missing something?

sirosen:
Ah, I apologize! I've been using `pre-commit` on several projects for a while, and I didn't properly consider that you might not be familiar with it.

As you say, I should add some note on it in the contrib doc -- I was going to do that later, but let's make it part of this PR. I'll update, hopefully this evening, with that.

Just to get you up to speed, here's a small brain-dump of the most relevant info about it:

Like tox, it's expected to be installed and available. I expect most users do `pipx install pre-commit` or `pip install --user pre-commit`.
You can opt-in to running it as a pre-commit hook by running `pre-commit install`, which will set it up as a hook in that repo. Or you can run it as a manual linter & fixer with `pre-commit run` (to run on files with staged changes) or `pre-commit run -a` to run on all files.

I also set it up as a `tox` environment -- a trick I picked up from some other projects -- so that even without `pre-commit` installed, you can `tox -e lint` to run it.

If you install a pre-commit hook, then a failing hook will abort a commit. So typically you would `git commit`, maybe it fails, you fix and add the necessary changes, and then you run `git commit` again.

If you ever have a broken pre-commit hook or need to force things through with it failing, you can turn it off by just running `git commit --no-verify`, which skips any hooks. Or you can remove the generated hook from `.git/hooks/` and run `pre-commit install` if you ever want it back.

I used to have to keep tox envs or make targets in a lot of projects which would run `flake8` or `isort` or whatever we wanted with any necessary arguments. `pre-commit` has largely replaced that and removed the drudgery of having to maintain a makefile which runs `.venv-linting/bin/pip install 'flake8==...' 'isort==...'` and so forth. At work I've found that it speeds things up by reducing the number of commits that actually go into a branch with linting failing, which just results in a bunch of `fix linting` commits and slower turnaround time between writing a change and having a passing build for a branch.

---

I'll try to translate the essential parts of the above about usage into something for the contrib doc.

stefanholek:
So it's installed via tox.ini, that is a clever trick indeed. 😃

sirosen:
Okay, it took me quite a bit longer than I expected to get an evening to work on this, but I've just put together a revision to the contrib doc. While I was in there, I tried to cut out a lot of the "excess" since I think the current contrib doc is too long and not focused enough.

PR #500, opened 2021-10-12
Refactor deprecated unittest aliases for Python 3.11 compatibility.

tirkarthi:
Deprecated unittest aliases were removed in Python 3.11 . This will cause error when tests are executed in Python 3.11 .

Ref : python/cpython#28268

### DIFF
--- a/nose2/plugins/loader/parameters.py
+++ b/nose2/plugins/loader/parameters.py
@@ -24,13 +24,13 @@ of the test function is the "case" argument, followed by the other parameters::
         @it.should('do bar')
         @params(1,2,3)
         def test(case, bar):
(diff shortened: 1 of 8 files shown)


sirosen:
Thanks for going the extra mile and opening issues and PRs for this! (Not just here, but I see on other projects as well.)

I'll get myself a py3.11 build setup for testing soon, but I don't see a strong reason to wait on that before merging.

tirkarthi:
Thanks @sirosen . In case you are using Ubuntu deadsnakes ppa can help with Python 3.11 builds https://launchpad.net/~deadsnakes/+archive/ubuntu/nightly

PR #505, opened 2021-12-10
add timestamp support to junitxml plugin

deeplow:
Per [spec definition](https://github.com/junit-team/junit5/blob/86b6b85/platform-tests/src/test/resources/jenkins-junit.xsd), junit-xml can have timestamp support (or the closest thing we have to a spec).
This adds that support, following the timestamp string definition according to the [date and time ISO 8601 standard](https://www.iso.org/iso-8601-date-and-time-format.html).

Regarding the choice of timestamp formatting, there doesn't seem to be an authoritative source on the spec, but in [this thread](https://stackoverflow.com/questions/442556/spec-for-junit-xml-output) people suggest Apache Ant project responsible for implementing the spec. In their implementation we see [the use of the ISO standard](https://github.com/apache/ant/blob/3a4980e/src/main/org/apache/tools/ant/taskdefs/optional/junit/XMLJUnitResultFormatter.java#L141):

```java
        //add the timestamp
        final String timestamp = DateUtils.format(new Date(),
                DateUtils.ISO8601_DATETIME_PATTERN);
        rootElement.setAttribute(TIMESTAMP, timestamp);
```

Pytest's junitxml implementation [also uses timestamps](https://docs.pytest.org/en/6.2.x/_modules/_pytest/junitxml.html) and their implementation (again) [uses the date and time ISO format](https://github.com/pytest-dev/pytest/blob/062d91a/src/_pytest/junitxml.py#L669).

### DIFF
--- a/nose2/plugins/junitxml.py
+++ b/nose2/plugins/junitxml.py
@@ -39,16 +39,16 @@ Produces this XML by default:
 
 .. code-block:: xml
 
-    <testcase classname="a" name="test_gen:1" time="0.000171">
+    <testcase classname="a" name="test_gen:1" time="0.000171" timestamp="2021-12-09T21:28:09.686611">
         <system-out />
     </testcase>
-    <testcase classname="a" name="test_gen:2" time="0.000202">
+    <testcase classname="a" name="test_gen:2" time="0.000202" timestamp="2021-12-09T21:28:09.686813">
         <system-out />
     </testcase>
-    <testcase classname="a" name="test_params:1" time="0.000159">
+    <testcase classname="a" name="test_params:1" time="0.000159" timestamp="2021-12-09T21:28:09.686972">
         <system-out />
     </testcase>
-    <testcase classname="a" name="test_params:2" time="0.000163">
+    <testcase classname="a" name="test_params:2" time="0.000163" timestamp="2021-12-09T21:28:09.687135">
         <system-out />
     </testcase>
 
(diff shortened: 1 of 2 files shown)


deeplow:
I have also made some tests in https://github.com/deeplow/nose2/commit/8751af36b1ab5930c8f81a2be62970c18aa82ba3. Let me know if I should add them here or as a separate PR.

sirosen:
Thanks for the contribution, and for the detailed background information!
This looks good to me, but I'd like the tests included in the PR.

I see that CI is broken right now. I don't get much time for nose2, so I suspect that something has drifted and there's a problem with `master`. I'm hoping I can get that sorted out and we can see a passing run here before I merge.

deeplow:
@sirosen I've added now the tests. Let me know if they are fine. Thanks.

deeplow:
We detected an instance where the test has no `.start` time ([see here](https://github.com/QubesOS/openqa-tests-qubesos/pull/16#issuecomment-999162166)). Not sure what conditions cause that to happen but should be elsewhere in the code unrelated to the plugin.

`[_time](https://github.com/nose-devs/nose2/blob/master/nose2/plugins/junitxml.py#L282)` in the plugin also handles this case by setting the time to 0. So my solution was to set the timestamp to 0 (which is jan 1st 1970 in unix time). Is this appropriate?

sirosen:
> I've added now the tests. Let me know if they are fine. Thanks.

I've been on a break, so I didn't get to look at this again until now. Sorry for the delay.
I've just approved the workflow run, so there's a [github actions run](https://github.com/nose-devs/nose2/runs/4695629825) to look at.

Everything looks good at a glance (I'll do a more thorough review in the next few days), but the run on the previous commit failed in github actions. Looking at the run, it seems that linting is failing.
You can run the linters and fixers with `tox -e lint`. If you can get that passing, that would help. (Worst case, I can circle back and update with fixes for linting; I don't mind. 😄 )

Once again, thanks for contributing!

> [_time](https://github.com/nose-devs/nose2/blob/master/nose2/plugins/junitxml.py#L282) in the plugin also handles this case by setting the time to 0. So my solution was to set the timestamp to 0 (which is jan 1st 1970 in unix time). Is this appropriate?

Yes, I think it's the best / most appropriate way to handle this situation.
The absence of a start time strikes me as a bug, unless there's something I've missed. But bug or not, the fact is that it happens. I think `nose2` should endeavor not to crash the user's testsuite unnecessarily, so this seems like an acceptable workaround.

deeplow:
> I've been on a break, so I didn't get to look at this again until now. Sorry for the delay.

No worries! Take your time.

> You can run the linters and fixers with `tox -e lint`. If you can get that passing, that would help. (Worst case, I can circle back and update with fixes for linting; I don't mind. smile )

Ran this now. The only issue is the docstrings that got over the 88 char limit due to the addition of the timestamps. I've broken it down into two lines. Not sure this is how you want to handle this, but it's something minor. Feel free to change it as you see fit :smile: .

sirosen:
I had to make some minor adjustments to get the tests passing. Mostly avoiding use of `datetime.fromisotimestamp` (because that method wasn't added until somewhere in the python 3.x series).

I also simplified the handling of `self._time` being `None` a little, but it's not a radical change.

This is good to go forward. I'll be taking some time to work on a release soon, so hopefully you'll be able to use it shortly! 🚀

deeplow:
Awesome! Thanks for the review!

PR #507, opened 2021-12-30
Add support for Python 3.10

hugovk:
Python 3.10 was released on 2021-10-04:

* https://discuss.python.org/t/python-3-10-0-is-now-available/10955

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -27,6 +27,7 @@ jobs:
           nose2 -v --pretty-assert
   test:
     strategy:
+      fail-fast: false
       matrix:
         # for the matrix, macos and linux on
         #   - py2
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -27,6 +27,7 @@ jobs:
           nose2 -v --pretty-assert
   test:
     strategy:
+      fail-fast: false
```
sirosen, line 30: Handy! I know it's somewhat OT, but thanks for sharing this. I'll have to use it in other projects.

.github/workflows/build.yaml
```diff
@@ -36,24 +37,26 @@ jobs:
         # any additional builds for windows and other python versions are
         # handled via `include` to avoid an over-large test matrix
         os: [ubuntu-latest, macos-latest]
-        python-version: [2.7, 3.5, 3.9]
+        python-version: ["2.7", "3.5", "3.10"]
```
sirosen, line 40: Thanks for converting everything to strings. Saves us the inevitable heartache when 3.20 rolls around! 😉

sirosen (approved):
Thanks for this, and all of the CI-related tidying-up!

PR #519, opened 2022-04-26
Fix for additional test method name on Python 3.11

hugovk:
Tests are failing on Python 3.11 ([for example](https://github.com/nose-devs/nose2/runs/6151776462?check_suite_focus=true)) because there was a change to unittest in March to add the fully qualified name to the output:

* https://github.com/python/cpython/issues/58473
* https://github.com/python/cpython/pull/32138

This PR fixes the nose2 tests to include the method name for Python 3.11 (and newer).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.orig
 example.cfg
 .coverage
+.idea
 nosetests.xml
 xunit.xml
 nose2-junit.xml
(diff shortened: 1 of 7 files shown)


sirosen:
Thanks for the fix! I'll take a look at this soon.

sirosen:
This fell off my radar, and I just caught it in a sweep over my review queue. Just merged.

cclauss:
Tests on [Py311 beta 4](https://www.python.org/download/pre-releases) are currently broken.

PR #523, opened 2022-07-15
Discover and fix typos with codespell

cclauss:
`codespell -w` to fix all typos. -- https://pypi.org/project/codespell

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,10 @@ repos:
     language_version: python3
     types: [python]
     exclude: ^nose2/_vendor/.*
+- repo: https://github.com/codespell-project/codespell
+  rev: v2.1.0
+  hooks:
+    - id: codespell
 
 
 ci:
--- a/docs/plugins/coverage.rst
+++ b/docs/plugins/coverage.rst
@@ -22,7 +22,7 @@ Otherwise, please be aware of the following known differences:
 Compatibility with mp plugin
 ----------------------------
 
-The ``coverage`` and ``mp`` plugins may be used in conjuction to enable
+The ``coverage`` and ``mp`` plugins may be used in conjunction to enable
 multiprocess testing with coverage reporting.
 
 Special instructions:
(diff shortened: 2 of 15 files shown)


.github/workflows/build.yaml
```diff
@@ -48,11 +68,11 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: install tox
-        run: python -m pip install -U tox
+        run: pip install tox
       - name: test
-        run: python -m tox -e py
+        run: tox -e py
```
sirosen, line 78: These invocations don't matter. In the context of GitHub Actions, `python -m tox` and `tox` are the same.
However, it's nicer netiquette (IMO) to leave things as they are unless there's a need to change them. I'd rather _not_ have to rethink the invocation of tox in a PR which isn't explicitly about or impacted by that.

.github/workflows/build.yaml
```diff
@@ -48,11 +68,11 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: install tox
-        run: python -m pip install -U tox
+        run: pip install tox
```
sirosen, line 76: Dropping the `-U`, in contrast with `python -m pip` vs `pip`, changes the meaning of this command.

Whether or not it's _wise_ to `pip install -U` in a build or not is something I'm happy to debate, but this is a big part of why I don't like seeing unrelated changes slipped into a PR. It's all too easy to accidentally change something which was done on purpose. Let's make sure we leave these lines alone in any refinement of this PR.

sirosen (changes requested):
I've never used `codespell` before but I'm interested in it already. It looks like it has the necessary level of configurability (mostly, ignore rules) to be safely usable. Thanks for bringing this up and proactively applying the changes!

However, there are some changes which are needed before this can merge. Primarily, I think it needs to be rebased to pick up the latest GitHub Actions config, and I've left some inline comments on a few things of note.

sirosen:
This looks good now, thanks!

If it ever becomes an issue for any reason, the linting can always be removed. And it fixes some very old typos, so that's great.

PR #524, opened 2022-07-15
Upgrade GitHub Actions

cclauss:
Failing test fixed in #529

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: 3.x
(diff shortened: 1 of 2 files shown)


hugovk (approved):
👍 

Maybe update `test-on-downstream-projects.yaml` as well?

sirosen:
I will note briefly that I don't like having additional little changes slipped in which aren't mentioned in the PR message or commit messages. It makes it harder to know what's intentional, or to have a good conversation about the changes. In this case, I'm referring to the updates to use `3.x` and `pypy-3.8` in the build.

That said, those changes are good anyway! Big thanks for the changes, and to @hugovk for stepping in to do review and feedback! 👍 

I'm also going to follow-up on this by setting up dependabot to do automatic PRs to update github actions.

PR #527, opened 2022-07-15
README.rst: Link to pypi stats

cclauss:
Failing test fixed in #529

https://pypistats.org/packages/nose2


sirosen:
This doesn't appear to do what it says it does? I don't mind adding a link to pypistats at all. It's a super-useful project. But I don't see such a link here.

cclauss:
The second graph shows a decreasing number of installations from legacy Python.

PR #526, opened 2022-07-15
GitHub Actions: Temporarily allow Python 3.11-dev to fail

cclauss:
Tests on [Py311 beta 4](https://www.python.org/download/pre-releases) are currently broken so let's not block unrelated pull requests.

NOTE: This pull request does NOT fix test runs on `Python 3.11 beta 4`.  It merely allows them to fail until they are fixed.

PR #528, opened 2022-07-15
Fix regex for Python 3.11 beta 4

hugovk:
Closes https://github.com/nose-devs/nose2/pull/526 as no longer needed.

Python 3.11 has started failing after beta has been released. See for example PR https://github.com/nose-devs/nose2/pull/527.

Fixes:

`    Regex didn't match: 'self.assertTrue\\(x\\)\n\\s+\\^+\nAssertionError: False is not true' not found in 'test_old_assertion (unittest_assertion.test_prettyassert_unittestassertion.TestFoo.test_old_assertion) ... FAIL\n\n======================================================================\nFAIL: test_old_assertion (unittest_assertion.test_prettyassert_unittestassertion.TestFoo.test_old_assertion)\n----------------------------------------------------------------------\nTraceback (most recent call last):\n  File "/private/tmp/nose2/nose2/tests/functional/support/scenario/pretty_asserts/unittest_assertion/test_prettyassert_unittestassertion.py", line 7, in test_old_assertion\n    self.assertTrue(x)\nAssertionError: False is not true\n\n----------------------------------------------------------------------\nRan 1 test in 0.000s\n\nFAILED (failures=1)\n'`

---

Adding formatting for clarity:

Regex didn't match: 

```
self.assertTrue\\(x\\)\n\\s+\\^+\nAssertionError: False is not true
```

not found in 

```
test_old_assertion (unittest_assertion.test_prettyassert_unittestassertion.TestFoo.test_old_assertion) ... FAIL

FAIL: test_old_assertion (unittest_assertion.test_prettyassert_unittestassertion.TestFoo.test_old_assertion)
  File "/private/tmp/nose2/nose2/tests/functional/support/scenario/pretty_asserts/unittest_assertion/test_prettyassert_unittestassertion.py", line 7, in test_old_assertion
    self.assertTrue(x)
AssertionError: False is not true

Ran 1 test in 0.000s\n\nFAILED (failures=1)

```

---

So it appears there's no longer an extra line between `self.assertTrue(x)` and `AssertionError: False is not true`

### DIFF
--- a/nose2/tests/functional/test_prettyassert.py
+++ b/nose2/tests/functional/test_prettyassert.py
@@ -183,7 +183,6 @@ class TestPrettyAsserts(FunctionalTestCase):
         # look for typical unittest output
         expect_lines = [
             r"self.assertTrue\(x\)",
-            r"\s+\^+",
             "AssertionError: False is not true",
         ]
         if not DEBUG_RANGES:


sirosen:
Thanks very much for the patch! I think there's a slight adjustment we need to make to the logic right after this line. See the commit which introduced this: https://github.com/nose-devs/nose2/commit/b66885260907d7d5c1e8565fd0a13b392eb0d342

So maybe we just revert that while changeset? I need to look at the 3.11 notes when I can to see what's going on with the debug ranges behavior.

hugovk:
Ah yes, reverting would probably be better, to simplify things.

PR #529, opened 2022-07-15
Revert "Update tests to pass on py3.11"

hugovk:
This reverts commit b66885260907d7d5c1e8565fd0a13b392eb0d342 as suggested in https://github.com/nose-devs/nose2/pull/528#issuecomment-1185776389.

Replaces and closes https://github.com/nose-devs/nose2/pull/528.

Closes https://github.com/nose-devs/nose2/pull/526 as no longer needed.

### DIFF
--- a/nose2/tests/functional/test_prettyassert.py
+++ b/nose2/tests/functional/test_prettyassert.py
@@ -1,18 +1,4 @@
-import sys
-
 from nose2.tests._common import FunctionalTestCase, windows_ci_skip
-
-# detect if DEBUG_RANGES are enabled in the interpreter
-#
-# for now, assume it's on for py3.11+
-#
-# in 3.11 this is be part of the env var flag loading logic:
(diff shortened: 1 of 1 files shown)


sirosen:
Thanks for this! I just did due diligence on 3.11.0b3 vs 3.11.0b4 and can confirm that this is a behavioral change on `self.assertTrue(False)` (for example) in unittest, and not just some weirdness injected by the nose2 tests.

Scanning/searching through the commits between those versions, I couldn't figure out the cause. But I'm not an expert in cpython's code, so it could easily be something "obvious" that I can't recognize.

So here's what I'm going to do:
- 🚀 merge this right now! 🚀 
- ❓ post on discuss.python.org to ask if the change is intentional or if I should expect that it could change again
- 🧠 remember to be ready to revert-revert if this changes back in a later 3.11 version

I'll start looking through the other open PRs as soon as I get that discourse post up.

sirosen:
For easy cross-linkage, here's the discourse post:
https://discuss.python.org/t/is-this-change-in-debug-ranges-between-3-11b3-and-3-11b4-intentional-expected/17409

PR #530, opened 2022-07-16
Provide clarity on `pip --upgrade` vs. `--user`, etc.

cclauss:
`pip install` has [command line options](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-user) for:
* --upgrade
* --upgrade-strategy
* --user
* --use-pep517

In a workflow, it is better to be explicit about which of the four we are using.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,6 +53,6 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: install tox
-        run: python -m pip install -U tox
+        run: python -m pip install --upgrade tox
       - name: test
         run: python -m tox -e py


sirosen:
`-U` is explicit and unambiguous.
```
$ pip install -h | grep '\-U'
  -U, --upgrade               Upgrade all specified packages to the newest
```

This is something about which I care _extremely_ little. But nothing is being fixed here. If someone doesn't know what `-U` does, they can look it up.

I'm drawing a line in the sand here. I'm not accepting this change. It seems like change for change's sake.

cclauss:
My sense is that we do not want users of node2 spending their time looking things up.  Explicit is better than implicit.

sirosen:
This is CI configuration. It is not user facing.
And it's not implicit. It's explicit.
Before quoting "Zen of Python", please take a moment to think about what the words mean.

Handling pointless PRs like this saps time and energy which could be better spent on material improvements to the software.

PR #531, opened 2022-07-16
Python is a brand name that should be capitalized

cclauss:
`Python` is a programming language while `python` is a kind of snake.
See its consistent use on https://www.python.org


sirosen:
I'm fine with this, but I make zero commitment to try to be consistent about it in the future.

If docs say `python2` somewhere, then I would hope that any user of python can figure out that this does not refer to a member of Serpentes but instead to a version of a popular programming language.

PR #532, opened 2022-07-16
flake8: Set a limit on cyclomatic code complexity

cclauss:
https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-max-complexity
* https://en.wikipedia.org/wiki/Cyclomatic_complexity
* https://github.com/nose-devs/nose2/pull/525#issuecomment-1186023584

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,6 +7,7 @@ known_third_party=six,coverage,mock
 
 [flake8]  # black and isort compatible rules
 extend-ignore = E203
+max-complexity = 17
 max-line-length = 88
 per-file-ignores =
     # allow use of `assert False` in tests -- it's fine


sirosen:
I'm okay with this in principle, but I'm not sure it's valuable to add this check in this project. This is a legacy codebase which has generally poor maintainability scores under maintainability metrics, so the threshold needs to be exceptionally high.
(Try `radon cc nose2 -n C` for a sample of places to look.)

And McCabe and Halstead are just proxies for code quality anyway...

---

It's fine to make this addition, but note that https://github.com/nose-devs/nose2/pull/525/commits/1dfbecb2149acb7578ba4969b533c9ed1a16724e moved it to the bottom of the config to keep it separate from the black-compatibility settings.

At the very least, please take a moment to not duplicate the commented-out config.

PR #549, opened 2022-11-05
build.yaml: Upgrade Python 3.11 production

cclauss:


cclauss:
I think of reproducibility as something that end users focus on but for CI, what we would really like to know (_before_ end users alert us) is when upgrades to current Python break the build.

sirosen:
FWIW, I think both things are true. We want CI to show us clear signal about issues related to new releases, but it's also good to be able to isolate that signal from the changes that you're working on.

If you happen to push a new branch or PR when something external to your changes has broken, it can be hard to de-noise that and identify the cause.

PR #568, opened 2023-04-19
Replace flake8, bugbear, isort, and pyupgrade with ruff

cclauss:
Fixes: #569

[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

The ruff Action uses minimal steps to run in ~5 seconds, rapidly providing intuitive GitHub Annotations to contributors.

image

Running the complete test matrix on GitHub Actions takes a lot of time and CPU.  Given that the `ruff` linting job is so fast (~5 seconds) let's ensure that linting passes ___before___ the complete test matrix is run.  This will cost a few seconds on each run but could save a lot of time and contention for PRs that should be cleaned up before they are fully tested.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds
Screenshot 2023-04-23 at 11 38 36


sirosen (approved):
I've been looking for a project to try out `ruff`, so putting it on `nose2` would be interesting to me.

I want to give particular thanks for being diligent about ignoring things which could have changed semantics, especially on old pythons. It make reviewing this much easier!

There's the possibility of moving `nose2`'s packaging configuration into `pyproject.toml` in the future to consider too.

Overall, a big :+1: for this. I look forward to seeing how it works out (and it can always be changed back in the future if necessary). Thanks!

Issue #569, opened 2023-04-19
GitHub tests are failing on Python 3.12 alpha 7


cclauss:
See https://github.com/nose-devs/nose2/actions

sirosen:
My thanks to both of you for running this down and providing the breadcrumbs for me to understand what was happening. I've just merged @cclauss's fix and will make sure to cite him in the changelog under py3.12 compatibility.

PR #570, opened 2023-04-23
Add addDuration() for Python >= 3.12

cclauss:
A subset of #568
Fixes: #569

See the Python 3.12 changes discussed at https://github.com/nose-devs/nose2/issues/569#issuecomment-1515828179

### DIFF
--- a/nose2/result.py
+++ b/nose2/result.py
@@ -50,6 +50,9 @@ class PluggableTestResult:
         event = events.StopTestEvent(test, self, time.time())
         self.session.hooks.stopTest(event)
 
+    def addDuration(self, test, elapsed):  # For Python >= 3.12
+        pass
+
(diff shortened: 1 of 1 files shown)


sirosen:
Thanks for contributing this! I'm going to read up a bit both on what we have in `nose2` and on how this is implemented in `unittest` before I merge, to confirm that making it a no-op is the right thing to do.

sirosen:
Okay, after reading through again, I don't think there's an obvious better quick solution. Probably this would translate to a new event type in nose2 if we wanted to continue to try to follow the model which exists, but I'm not sure that's such a great idea given the current age and status of the project (and how little time I end up making for it).

Once again, thanks for not only identifying this but also putting in a PR with the fix.

PR #571, opened 2023-04-25
tox.ini: allowlist_externals

cclauss:
Fix for https://github.com/tox-dev/tox/issues/2599


sirosen:
Thanks! I might also set `minversion` soon, since I'll probably use a tox4 feature at some point by accident.

PR #573, opened 2023-04-25
Fix issues with license data

### LINKED ISSUE #551: sdist: PKG-INFO should list license.txt in License-File
The sdist at PyPI contains this:
```
$ grep -i license nose2-0.12.0/PKG-INFO
Classifier: License :: OSI Approved :: BSD License
License-File: AUTHORS
$
```
But the actual `nose2` license is in `license.txt` file.  Please add `license.txt` to `PKG-INFO`.

Thank you.

### LINKED ISSUE #553: License terms for unittest2?
The license file indicates that some of the code is derived from unittest2:

https://github.com/nose-devs/nose2/blob/5375b1d5eee5fa59a1785098fedc4441d6cfcb55/license.txt#L30-L32

The link http://docs.python.org/license.html does not make it clear what license terms are intended here, and I’m not sure that this is the right link anyway since unittest2 was never part of the standard library.

The [pypi package for `unittest2`](https://pypi.org/project/unittest2/) has a “BSD” license trove classifier, and its `setup.py` references http://www.voidspace.org.uk/python/license.shtml, a now-dead link that [previously hosted “The Voidspace Open Source License” (Wayback Machine)](http://web.archive.org/web/20200709010030/http://www.voidspace.org.uk/python/license.shtml), which was a [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause.html) license. I think that license text is what should apply here.

If you believe these are the correct terms for unittest2 code in nose2, would you consider updating the license file with the actual license text?

sirosen:
1. `license.txt` is not part of the default set of files which setuptools uses for license info, but `AUTHORS` is, resulting in setuptools picking the wrong file to act as the license.
2. License info for unittest2 was incorrect and incomplete. Reproduce the license info verbatim here.

For (1), rename the license file to `LICENSE`. This makes it more conventional and fixes the setuptools issue.

For (2), pull in the original unittest2 license and copyright info by examining the unittest2 repo.

fixes #551, fixes #553


musicinmybrain:
Thanks for dealing with this!

PR #580, opened 2023-08-30
Update config to drop py3.7 and add 3.12

sirosen:
Update configs to add 3.12.0rc1 to build matrix.

Also update requires-python data.

---

The 3.12 builds have been broken on some of the alphas; a local test on 3.12.0rc1 worked for me, so try it out in GitHub.


sirosen:
Yeesh. Not loving that there's a CI run of ruff which is not bound to the same version as pre-commit. I've removed that.

PR #596, opened 2023-12-27
[Feature] Add support for toml

### LINKED ISSUE #452: Support configuring through pyproject.toml
The [`tool` table section](https://www.python.org/dev/peps/pep-0518/#tool-table) of PEP 518 says that tools should store their name in `[tool.<pypiname>]` in `pyproject.toml`.

qequ:
Fixes #452


JCHacking:
Now in python 3.12 there is a built-in library to interact with toml, it could be applied to avoid installing this dependency.

sirosen:
I haven't been making enough time for nose2 lately, but this PR currently isn't quite in a state that I feel I can merge.
I've needed to take time to offer feedback, since it's _close_, as far as I'm concerned.

- it needs to be restricted to `pyproject.toml`, not any toml file
- I think we should be using `tomli` on py<3.11 and stdlib `tomllib` on 3.11+ -- partly because tomli is what was portted into tomllib
- :point_up: the dependency on `tomli` should be conditional on a python_version marker

Those are the main things.

I'm also a little bit concerned in theory that `nose2` could be confused and made to crash with certain inputs (namely, TOML-native datetime data, which would manifest as unexpected `datetime.datetime` objects in the config), but I don't know that the config loading which exists today is sufficiently hardened against these sorts of usages. So I wouldn't worry much about it but it would be nice to check on that, maybe after this or a similar changeset merges.

sirosen:
With #606 merged, I'm closing this.

I'll release in the next couple of days.

PR #606, opened 2024-05-08
Support loading config from pyproject.toml

sirosen:
This is a continuation of the work started in #596 -- thanks to @qequ for getting this started!

I've made some changes which I believe are important to nose2 supporting the full breadth of use cases, including test matrices which run tests with TOML parsers present and absent in lower python versions.
In general, I think `nose2` should avoid addig dependencies outside of the stdlib, since it is a test runner.
There's some room for healthy debate on this, but with the current course of `tomli`->`tomllib`, I don't think it's necessary for TOML support to add a hard dependency or an extra.


sirosen:
I finally made a few minutes to look into those CI failures. The cause was pretty simple: missing `*.toml` from the `MANIFEST.in` includes. I've rebased + added the relevant tweak.

PR #608, opened 2024-05-09
Adjust the regex in test_unittest_assertion for Python 3.13

musicinmybrain:
Before this PR, one test fails on Python 3.13 (3.13.0a6):

```
FAIL: test_unittest_assertion (nose2.tests.functional.test_prettyassert.TestPrettyAsserts.test_unittest_assertion)
  File "/home/ben/src/forks/nose2/nose2/tests/functional/test_prettyassert.py", line 171, in test_unittest_assertion
    stderr = self.assertProcOutputPattern(proc, expected)
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/ben/src/forks/nose2/nose2/tests/functional/test_prettyassert.py", line 15, in assertProcOutputPattern
    self.assertTestRunOutputMatches(proc, stderr=expected)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ben/src/forks/nose2/nose2/tests/_common.py", line 82, in assertTestRunOutputMatches
    self.assertRegex(util.safe_decode(cmd_stderr), stderr)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Regex didn't match: 'self.assertTrue\\(x\\)\nAssertionError: False is not true' not found in 'test_old_assertion (unittest_assertion.test_prettyassert_unittestassertion.TestFoo.test_old_assertion) ... FAIL\n\n======================================================================\nFAIL: test_old_assertion (unittest_assertion.test_prettyassert_unittestassertion.TestFoo.test_old_assertion)\n----------------------------------------------------------------------\nTraceback (most recent call last):\n  File "/home/ben/src/forks/nose2/nose2/tests/functional/support/scenario/pretty_asserts/unittest_assertion/test_prettyassert_unittestassertion.py", line 7, in test_old_assertion\n    self.assertTrue(x)\n    ~~~~~~~~~~~~~~~^^^\nAssertionError: False is not true\n\n----------------------------------------------------------------------\nRan 1 test in 0.000s\n\nFAILED (failures=1)\n'
```

Close inspection shows that Python 3.13 has inserted the line `    ~~~~~~~~~~~~~~~^^^` between `    self.assertTrue(x)` and `AssertionError: False is not true` to indicate the error location as part of https://docs.python.org/3.13/whatsnew/3.13.html#improved-error-messages.

This PR adds `(\W+\n)?` in the regex to accept one optional line composed entirely of non-word (non-alphanumeric) characters in the expected output.

After this PR, all tests pass with `tox -e python313-nocov`.

### DIFF
--- a/nose2/tests/functional/test_prettyassert.py
+++ b/nose2/tests/functional/test_prettyassert.py
@@ -167,7 +167,7 @@ class TestPrettyAsserts(FunctionalTestCase):
             "scenario/pretty_asserts/unittest_assertion", "-v", "--pretty-assert"
         )
         # look for typical unittest output
-        expected = "self.assertTrue\\(x\\)\nAssertionError: False is not true"
+        expected = "self.assertTrue\\(x\\)\n(\W+\n)?AssertionError: False is not true"
         stderr = self.assertProcOutputPattern(proc, expected)
         # the assertion line wasn't reprinted by prettyassert
         self.assertNotIn(">>> self.assertTrue", stderr)


sirosen:
Thanks for this! I think the addition to the regex is as simple as we could hope, so I'm going to merge right away. 

Do you need this to be released, or would it be enough to have it in main? I ask because I'm working on some stuff for a likely upcoming release, but I can put out a quick patch if it would be helpful. 

Also, I really need to get 3.13 testing added to tox and CI so that I can stay ahead of / on top of these issues. That should be a pretty easy fast-follow here.

musicinmybrain:
> Thanks for this! I think the addition to the regex is as simple as we could hope, so I'm going to merge right away.
> 
> Do you need this to be released, or would it be enough to have it in main? I ask because I'm working on some stuff for a likely upcoming release, but I can put out a quick patch if it would be helpful.

I’m not *personally* in a big hurry to have it in a release. It’s easy enough to carry a downstream patch in [the Fedora package](https://src.fedoraproject.org/rpms/python-nose2) for now so we can proceed with [Python 3.13 testing](https://copr.fedorainfracloud.org/coprs/g/python/python3.13/) of the packages that use `nose2`.

To help you plan, here’s the Python 3.13 release schedule: https://peps.python.org/pep-0719/

> Also, I really need to get 3.13 testing added to tox and CI so that I can stay ahead of / on top of these issues. That should be a pretty easy fast-follow here.

Awesome! Thanks for working on this.

PR #613, opened 2024-05-31
Update _toml.py

grhwalls:
I think TOML_ENABLED should be set to True when tomli is found (when falling back from tomllib)

### DIFF
--- a/nose2/_toml.py
+++ b/nose2/_toml.py
@@ -8,7 +8,7 @@ except ImportError:
     try:
         import tomli as toml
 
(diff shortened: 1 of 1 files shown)


sirosen:
Good catch; you're 100% right! I'll make sure you get a shout-out in the changelog; thanks for the fix!

Between this and the config loading issue you reported, I think a v0.15.1 release will be right around the corner. I'll wait for CI on this; I don't want to push the fix out until at least trying to fix the config loading problem though, so it may take a little bit of time.

PR #616, opened 2024-06-05
updated fail-fast plugin to take 'n' as number of failures to stop the run

gautamkhanapuri:
configuration section for this change
[failfast]
failures=2

command line
nose2 -v -F 2
nose2 -v --fail-fast 4

### DIFF
--- a/nose2/plugins/failfast.py
+++ b/nose2/plugins/failfast.py
@@ -15,18 +15,47 @@ __unittest = True
 class FailFast(events.Plugin):
     """Stop the test run after error or failure"""
 
-    commandLineSwitch = (
-        "F",
-        "fail-fast",
-        "Stop the test run after the first error or failure",
-    )
+    configSection = "failfast"
+
+    def __init__(self):
+        """Initialize failfast as stop and failcount as current count"""
+        self.failfast = 0
+        self.failcount = 0
+        self.addArgument(
+            self.setFailFast,
+            "F",
+            "fail-fast",
+            "Stop the test run after the first error or failure (1 = auto)",
+        )
+        # Number of failures to stop the execution
+        self.failfast = self.config.as_int("failures", 0)
+
+    def handleArgs(self, args):
+        """Register if fail-fast is >0 in cfg or command-line"""
+        if self.failfast > 0:
+            self.register()
+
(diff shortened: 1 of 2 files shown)


sirosen:
Hi there, thanks for this contribution! (Sorry it's taken so long to get you feedback; it's been in my queue for a while but it's hard for me to make time for all of my various projects.)

I think fundamentally the idea here is sound -- pytest's failfast behavior also supports this kind of usage -- but I see two issues with the way that it's done here. If you can rectify these, I can see merging this change as a new feature and doing a release.

## 1. Breaking Interface Changes

I don't believe that changing the meaning of an existing flag is an acceptable change for `nose2` at this stage of its lifecycle without very, very strong motivation.

`--fail-fast` is a boolean flag, and it should remain one.

I would strongly encourage you to take the following approach instead:
- `--max-fail N` is a new option which controls the allowed number of failures.
- `--fail-fast` is a flag which, under the default behavior, means that `--max-fail 1` has been set
- rules are established for how `--max-fail` and `--fail-fast` interact

Regarding the rules of interaction, there are a lot of possible modes, none of which should be terribly surprising, especially if they're added to the fail-fast plugin docs. For example, the flags could be treated as mutex, or `--max-fail 1` could _imply_ `--fail-fast`, or the combination could be treated as necessary for `--max-fail` to be effective. I don't feel strongly about prescribing such a behavior -- I think `--max-fail` implying `--fail-fast` but the two being compatible is nicest, but it might be harder to implement -- but it has to be well-defined.

The important issue here is the interface breakage. Because the primary user-base for `nose2` are legacy applications, rather than new ones, breaking changes should be avoided whenever it is feasible.

## 2. New tests

Any significant functionality change like this needs new tests which exercise the feature.
I'm happy to help with this as I'm able (though, as demonstrated, it could take me quite a bit of time to get to it), but I'm not going to try to add any right now. Without the interface being sorted out, it's probably not a good use of time.

---

Thanks again for coming to the table with this contribution! I'm hopeful that we can resolve these and get this into an upcoming release!

sirosen:
Given the time without a response here, I'm closing. 
This feature would be perfectly welcome so long as it's done in a safe and compatible way.

PR #644, opened 2026-01-07
plugins/mp: set context to fork for Python 3.14 mp API changes

JimmyDurandWesolowski:
This commit sets the Python built-in multiprocessing library context to multiprocessing.context.ForkContext.
This is needed due to a change in its API from Python 3.14:
> Changed in version 3.14: On POSIX platforms the default start method was
> changed from fork to forkserver to retain the performance but avoid
> common multithreaded process incompatibilities. See gh-84559.

Note that the change is done in the "mp" module itself and not in its MultiProcess class since the multiprocessing context can be set only once.

### DIFF
--- a/nose2/plugins/mp.py
+++ b/nose2/plugins/mp.py
@@ -15,6 +15,17 @@ from nose2 import events, loader, result, runner, session, util
 log = logging.getLogger(__name__)
 
 
+# Python 3.14: Change in multiprocessing API
+# https://docs.python.org/3/library/multiprocessing.html
+# #multiprocessing-start-methods
+# On POSIX platforms the default start method was changed from fork to
+# forkserver to retain the performance but avoid common multithreaded
+# process incompatibilities.
+# [...]
+# set_start_method() should not be used more than once in the program.
+multiprocessing.set_start_method("fork")
+
+
 class MultiProcess(events.Plugin):
     configSection = "multiprocess"
 


airtower-luna:
This causes trouble (specifically: `RuntimeError: context has already been set`) with any code that also uses `multiprocessing.set_start_method()`. Wouldn't it make more sense to create a multiprocessing context for nose (or the mp plugin), so the required start method can be used without touching global settings?

See:
* https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_context
* https://docs.python.org/3/library/multiprocessing.html#multiprocessing-start-methods

sirosen:
I missed this PR, sorry! 

Yes, the change here is not correct. I think using the context API is the right approach. I'd be happy to review a PR with a patch for it. We're also missing 3.14 from CI, which probably needs to happen first.

airtower-luna:
I've experimented a bit, there seem to be two things that break with different multiprocessing start methods:

1. This one because help output ends up colorized (extra characters), easily fixed by adding `PYTHON_COLORS=0` to `setenv` in `tox.ini`: https://github.com/nose-devs/nose2/blob/e400806e6f6f1b707103d31c545bb478590ae96c/nose2/tests/unit/test_plugin_api.py#L18-L22
2. Tests with coverage enabled often produce "No data to report" errors, while the matching `py*-nocov` runs are perfectly fine (pass!) after fixing the color issue.

Because of 2.) I suspect changing the multiprocessing start method isn't actually the correct fix, and instead the coverage setup/configuration needs to be fixed. Also, setting the start method to `fork` unconditionally breaks on Windows (because it's not available there), and considered "unsafe" on macOS according to `multiprocessing` documentation. Both default to `spawn`.

I've pushed my experiment here: https://github.com/airtower-luna/nose2/tree/python-3.14-fix

You can see the Windows breakage in the test results, and also MyPy errors because `pkg_resources` is missing (no surprise, it had been [deprecated for a long time](https://setuptools.pypa.io/en/latest/deprecated/pkg_resources.html) and has now been removed with setuptools v82.0.0).

Commit https://github.com/airtower-luna/nose2/commit/291ea09f2c025578e4fb50a0d8e71af1b06a069a could still be useful as a less-breaking patch for distros, or as a starting point for a workaround. I don't have time to dig into the coverage issues for the foreseeable future.

sirosen:
Thanks for kick-starting the investigation! I've started looking into this further and made a little bit of progress on my own (visible in [a branch here](https://github.com/nose-devs/nose2/tree/test-on-py314)). The colorized output was easy enough to fix with `NO_COLOR=1`.

I'm a bit nervous about these coverage-related test failures. The coverage plugin tests have always been a little flaky and weird. Although it wouldn't be my favorite way to resolve, that test might need to be removed on 3.14+...

I think we want the context work as well. That seems correct to me -- and it needs to be conditional per platform. I plan to pull that in on my branch as a co-authored commit when I circle back on this later in the week.

sirosen:
@airtower-luna, I'm picking up a cherry-pick of your custom context commit and putting it into a PR right now!
I'll make sure to give a shout-out to you in the changelog.

I realized I've been wanting to get rid of the `coverage` plugin because it's both flaky to maintain and unnecessary (using `coverage` to invoke `nose2` is better in many ways), and it keeps me from doing maintenance on `nose2` that I might otherwise be perfectly happy to spend time doing.
#651 details the deprecation plan, which includes not testing it at all on py3.14.

sirosen:
closed via #655

PR #677, opened 2026-07-25
fix: avoid UnboundLocalError when a plugin clears event.names

### LINKED ISSUE #445: suites not always assigned in loadTestsFromNames in loader.py
If a plugin implements the `loadTestsFromNames` hook, doesn't set `event.handled=True` and sets `events.names` to something Falsey (e.g. []), suites is never assigned.  
I think there could be a scenario in which a plugin would want to remove certain or all test names from events.names without setting handled=True, and that suites should default to an empty list in this case.

Example plugin (run with any test):

```python
class Example(Plugin):
    alwaysOn = True

    def loadTestsFromNames(self, event):
        event.names = []
```

Trace:

```
  File ".../bin/nose2", line 10, in <module>
    sys.exit(discover())
  File ".../lib/python3.7/site-packages/nose2/main.py", line 301, in discover
    return main(*args, **kwargs)
  File ".../lib/python3.7/site-packages/nose2/main.py", line 87, in __init__
    super(PluggableTestProgram, self).__init__(**kw)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/main.py", line 100, in __init__
    self.parseArgs(argv)
  File ".../lib/python3.7/site-packages/nose2/main.py", line 120, in parseArgs
    self.createTests()
  File ".../lib/python3.7/site-packages/nose2/main.py", line 253, in createTests
    self.testNames, self.module)
  File ".../lib/python3.7/site-packages/nose2/loader.py", line 73, in loadTestsFromNames
    return self.suiteClass(suites)
UnboundLocalError: local variable 'suites' referenced before assignment
```

[Relevant method in loader.py](https://github.com/nose-devs/nose2/blob/fd5af12b40dfa8d0bb660f415134fd943d1d9fdf/nose2/loader.py#L53-L73):

```python
def loadTestsFromNames(self, testNames, module=None):
        """Load tests from test names.

        Fires :func:`loadTestsFromNames` hook.

        """
        event = events.LoadFromNamesEvent(
            self, testNames, module)
        result = self.session.hooks.loadTestsFromNames(event)
        log.debug('loadTestsFromNames event %s result %s', event, result)
        if event.handled:
            suites = result or []
        else:
            if event.names:
                suites = [self.loadTestsFromName(name, module)
                          for name in event.names]
            elif module:
                suites = self.loadTestsFromModule(module)
        if event.extraTests:
            suites.extend(event.extraTests)
        return self.suiteClass(suites)
```

Sanjays2402:
Closes #445

If a `loadTestsFromNames` plugin clears `event.names` without setting `event.handled`, neither branch in `PluggableTestLoader.loadTestsFromNames` assigns `suites`, so the call raises `UnboundLocalError: local variable 'suites' referenced before assignment`. Defaulting `suites` to an empty list loads an empty suite instead, which is the behaviour the issue asks for.

Added a regression test registering a plugin that clears `event.names`; it fails with the reported `UnboundLocalError` without the loader change and passes with it.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,6 +19,10 @@ Fixed
 * JUnit XML reports now use the containing test's start timestamp for every
   failing subtest. (:issue:`567`)
 
+* ``PluggableTestLoader.loadTestsFromNames`` no longer raises
+  ``UnboundLocalError`` when a plugin clears ``event.names`` without setting
+  ``event.handled``. An empty suite is loaded instead. (:issue:`445`)
+
 0.16.0 (2026-03-01)
 -------------------
 
--- a/nose2/loader.py
+++ b/nose2/loader.py
@@ -64,6 +64,8 @@ class PluggableTestLoader:
                 suites = [self.loadTestsFromName(name, module) for name in event.names]
             elif module:
                 suites = self.loadTestsFromModule(module)
+            else:
+                suites = []
         if event.extraTests:
             suites.extend(event.extraTests)
         return self.suiteClass(suites)
--- a/nose2/tests/unit/test_loader.py
+++ b/nose2/tests/unit/test_loader.py
@@ -74,6 +74,13 @@ class TestPluggableTestLoader(TestCase):
             "FakePlugin.loadTestsFromNames() was not called",
         )
 
+    def test_loader_from_names_without_names_or_module(self):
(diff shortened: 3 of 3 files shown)


sirosen:
Thanks for this; it's a good fix.

Issue #373, opened 2017-11-16
SkipTest raised inside setupClass treated as error


Link512:
virtualenv python version: `3.6.2`
nose2 version: `0.7.2`
unittest2 version: `1.1.0`

snippet:
```python
import unittest2

class Failing(unittest2.TestCase):

    @classmethod
    def setUpClass(cls):

        raise unittest2.SkipTest

    def testWillBeIgnored(self):

        self.assertTrue(1)
```

Expected behaviour: Test case is skipped
Actual behaviour: SkipTest is treated like actual error. See below:
```
nose2
E
======================================================================
ERROR: _ErrorHolder
----------------------------------------------------------------------
Traceback (most recent call last):
  File "[...]/test_nose/test_failure.py", line 8, in setUpClass
    raise unittest2.SkipTest
unittest2.case.SkipTest

----------------------------------------------------------------------
Ran 0 tests in 0.000s
```

katrinabrock:
Thanks for reporting. I assume this worked fine under 0.6.5? Maybe pin there for now.

This functionality should get tested here:
https://github.com/nose-devs/nose2/blob/7928e015873a7dc22eb5d0693ae34a52dd8141f8/nose2/tests/functional/support/scenario/tests_in_package/pkg1/test/test_things.py#L23
(but with unittest not unittest2)

Link512:
Yes, with 0.6.5 it works fine.

katrinabrock:
Able to reproduce on python version 2.7

katrinabrock:
Unsurprisingly, this is where the bug was introduced:
https://github.com/nose-devs/nose2/commit/19822c9845ba37fa51e0bc29016c57254957d968

ptthiem:
'''python
// nose2.compat: :giggle:
import unittest
try:
    import unittest2
    SkipTestExcs = (unittest2.SkipTest, unittest.SkipType)
except ImportError
    SkipTestExcs = (unittest.SkipType, )
'''

(Hides)

sirosen:
I'm closing this as an issue about unittest2, which is therefore no longer relevant for the project.

Although I won't go out of my way to break unittest2 support, it's no longer a target. We don't install and test against it, and it has very likely drifted out of support, as it hasn't done any releases in the last 10 years.

PR #678, opened 2026-07-25
test: cover SkipTest raised in setUpClass

Sanjays2402:
Related: #373

## What this is

Regression coverage, not a fix. `SkipTest` raised in `setUpClass` already behaves correctly on `main` — I verified it before writing anything:

```
$ cat test_skipcls.py
import unittest

class Failing(unittest.TestCase):
    @classmethod
    def setUpClass(cls):
        raise unittest.SkipTest("skip whole class")

    def test_ignored(self):
        self.assertTrue(1)

$ python -m nose2 -v
_ErrorHolder ... skipped skip whole class
Ran 0 tests in 0.000s
OK (skipped=1)          # exit 0
```

The reported failure mode (the class reported as an `ERROR` via `_ErrorHolder`, exit non-zero) does not occur on 0.16.0.

What was never done is the thing @katrinabrock asked for in the issue thread:

> This functionality should get tested here: `.../scenario/tests_in_package/pkg1/test/test_things.py` (but with unittest not unittest2)

That scenario has `test_skippy`, a skipping *test method*, but nothing covering a skip raised from `setUpClass`. So the behavior is correct and completely unpinned, which is presumably how the issue stayed open while the underlying bug quietly got fixed.

## Approach

I put the fixture in a new `scenario/skip_in_class_setup/` rather than adding to `tests_in_package` as suggested. Six functional modules load that scenario and several assert on exact counts (`Ran 25 tests`), so adding classes there means updating unrelated expectations in the same PR. A dedicated scenario keeps the diff additive.

The fixture asserts the negative too: both methods in the skipped class raise `AssertionError` if they ever execute, so a regression that runs tests after a class-level skip fails loudly instead of just changing a count. A sibling class in the same module verifies the skip does not leak to unrelated tests.

## Testing

`nose2/tests/functional/test_loading.py`: 37 passed with this change, 36 on `main`. The new test asserts `OK (skipped=1)`, `Ran 1 test`, the skip reason text, and exit status 0.

No changelog entry: this adds no user-facing behavior change.

### DIFF
--- a/nose2/tests/functional/support/scenario/skip_in_class_setup/test_skip_in_setupclass.py
+++ b/nose2/tests/functional/support/scenario/skip_in_class_setup/test_skip_in_setupclass.py
@@ -1,0 +1,26 @@
+import unittest
+
+
+class SkipInSetUpClass(unittest.TestCase):
+    """SkipTest from setUpClass should skip the class, not error.
+
+    Regression coverage for issue #373: this was reported as an error
+    (an ``_ErrorHolder`` failure) rather than a skip.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        raise unittest.SkipTest("skipping the whole class")
+
+    def test_not_run_1(self):
+        raise AssertionError("setUpClass skipped; this must never run")
+
+    def test_not_run_2(self):
+        raise AssertionError("setUpClass skipped; this must never run")
+
+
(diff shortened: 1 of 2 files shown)


sirosen:
You found an old issue which can be closed and sent a slop PR instead. 

The issue was about unittest2 compatibility, which is no longer really relevant.

